PR #8, opened 2017-08-18
Add preferences editing inlined extension function

mitrejcevski:


app/src/main/java/Extensions.kt
```diff
@@ -57,4 +58,10 @@ fun MenuItem.loadIconFromUrl(context: Context, imageUrl: String) {
                     icon = circularIcon
                 }
             })
+}
+
+inline fun SharedPreferences.edit(f: SharedPreferences.Editor.() -> Unit) {
+    val editor = edit()
+    editor.f()
+    editor.apply()
```
chrisbanes: This needs to be `commit()`, since I let RxJava handle the threading. Maybe add a parameter to switch between `apply()` and `commit()`?

mitrejcevski: @chrisbanes Sure, will do. I was also thinking about splitting the `Extensions.kt` into package with extensions file per related topic (like views, preferences, etc). Or you prefer to have this one file with all extensions in it? What do you think?

chrisbanes: Separate files sounds good: `ViewExtensions`, `PreferenceExtensions`, etc. 👍

PR #16, opened 2017-08-23
Add 'More' label to list header

christxph:
Makes it clearer for users that the header can be clicked and looks similar to how the Google Play Store app does that.

### DIFF
--- a/app/src/main/res/layout/header_item.xml
+++ b/app/src/main/res/layout/header_item.xml
@@ -24,21 +24,22 @@
 
     <TextView
         android:id="@+id/header_title"
-        android:layout_width="match_parent"
+        android:layout_width="wrap_content"
         android:layout_height="match_parent"
-        android:layout_marginStart="14dp"
-        android:layout_marginEnd="14dp"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/header_item.xml
```diff
@@ -32,4 +32,16 @@
         android:gravity="center_vertical"
         tools:text="Game of Thrones" />
 
+    <TextView
+        android:id="@+id/header_more"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:layout_marginEnd="14dp"
```
piotrek1543: I wonder if you could extract hardcoded value `14dp` to dimens variable or use one similar which already exists. I suppose also that there is already an item with value `16dp`.

It would help others to easier manage code if there would be more similar views like this.

christxph: There aren't any `14dp` or `16dp` dimens used yet (except in this file), but I'll put put them into a `dimens.xml` for future uses

PR #17, opened 2017-08-24
Scroll to top when bottom navigation is clicked

christxph:
Scrolls the `RecyclerView` to the top when the bottom navigation tab is clicked and the backstack is empty. There should be a better way to do this than using `supportFragmentManager.fragments[1]` to get the current fragment.

### DIFF
--- a/app/src/main/java/me/banes/chris/tivi/home/HomeActivity.kt
+++ b/app/src/main/java/me/banes/chris/tivi/home/HomeActivity.kt
@@ -63,7 +63,11 @@ class HomeActivity : TiviActivity() {
                             supportFragmentManager.popBackStackImmediate()
                         }
                     } else {
-                        // TODO scroll to top of main fragment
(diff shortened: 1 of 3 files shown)


app/src/main/java/me/banes/chris/tivi/home/HomeActivity.kt
```diff
@@ -63,7 +63,11 @@ class HomeActivity : TiviActivity() {
                             supportFragmentManager.popBackStackImmediate()
                         }
                     } else {
-                        // TODO scroll to top of main fragment
+                        val fragment = supportFragmentManager.fragments[1]
```
chrisbanes: I'm not sure this is the correct way to fetch this. `findFragmentById(R.id.home_content)` is more appropriate.

christxph: Thanks, the other way would break sooner or later

PR #97, opened 2018-01-01
Reduce the number of classes named "*Manager"

vlazzle:
AppManager -> AppInitializer
TmdbManager -> TmdbImageProviderRepo

### DIFF
--- a/tmdb/src/main/java/me/banes/chris/tivi/tmdb/TmdbImageProviderRepo.kt
+++ b/tmdb/src/main/java/me/banes/chris/tivi/tmdb/TmdbImageProviderRepo.kt
@@ -30,7 +30,7 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class TmdbManager @Inject constructor(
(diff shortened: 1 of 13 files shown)


tmdb/src/main/java/me/banes/chris/tivi/tmdb/TmdbImageProviderRepo.kt
```diff
@@ -30,7 +30,7 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class TmdbManager @Inject constructor(
+class TmdbImageProviderRepo @Inject constructor(
```
chrisbanes: This should stay as `TmdbManager`. While it only currently provides the TmdbImageProvider, it will do more in the future (possible TMDb login, etc).

vlazzle: done

chrisbanes:
Thanks!

PR #105, opened 2018-01-10
Remove theme setting

chrisbanes:


app/src/main/java/me/banes/chris/tivi/TiviFragment.kt
```diff
@@ -55,7 +55,7 @@ abstract class TiviFragment : DaggerFragment() {
 
         if (postponed && !startedTransition) {
             // If we're postponed and haven't started a transition yet, we'll delay for a max of 200ms
```
vanniktech: That comment is now outdated.

chrisbanes: Woops, that was just for debug. Thanks for catching this.

PR #110, opened 2018-01-20
Remove sdk licenses in circleci config

DanielMartinus:
I wasn't sure if I should open a Pull Request for this. But I realized and read in slack groups that other developers were also looking at the CircleCI config of this project in order to build their own.

With a change in the Android DockerFile it's not longer necessary to update the sdk licenses in the config file. 
https://github.com/circleci/circleci-images/blob/master/android/Dockerfile.m4#L69

I've tested it out and removed it too with success. Unless you need it for something else that I'm not aware of.

As last I want to point out that Benoît Quenaudon made me aware of this.


chrisbanes:
Thanks, but I'm going to the update check. I've had situations in the past when the docker image had not been updated.

Since the check only takes ~1 second, there's no harm in keeping it in.

PR #129, opened 2018-04-01
Move away from jobs for updating shows

chrisbanes:
Due to serial nature of JobIntentService. Also completed a LARGE refactor.


app/src/main/java/me/banes/chris/tivi/inject/AppModule.kt
```diff
@@ -45,7 +48,12 @@ class AppModule {
     @Singleton
     @Provides
     fun provideRxSchedulers(): AppRxSchedulers {
-        return AppRxSchedulers()
+        return AppRxSchedulers(
+                Schedulers.single(),
+                Schedulers.io(),
+                Schedulers.io(),
```
jshvarts: are these duplicate io() params intended?

chrisbanes: Yep, they're used for different things.

PR #166, opened 2018-06-13
Interactor tests

chrisbanes:


app/src/androidTest/java/app/tivi/interactors/FollowInteractorTest.kt
```diff
@@ -0,0 +1,58 @@
...
+        followShow = FollowShowInteractor(testCoroutineDispatchers, followShowsDao, seasonFetcher)
+    }
+
+    @Test
+    fun test_doWork() {
+        runBlocking {
+            followShow(showId)
+
+            assertThat(followShowsDao.entryWithShowId(showId), `is`(notNullValue()))
+            verify(seasonFetcher, times(1)).load(showId)
```
jshvarts: just a matter of style but fyi: `times(1)` is a default so can be omitted.

chrisbanes: Yep, I prefer being explicit for this.

jshvarts: I thought so 👍

PR #173, opened 2018-06-26
Image loading improvements

chrisbanes:
Blatantly stealing code from @nickbutcher for the 'saturate on load' animator. Reimplemented it as a GlideType which makes it easier to use, and in Kotlin.


nickbutcher:
Neat!

PR #212, opened 2018-08-30
MvRx explorations WIP

chrisbanes:


app/src/main/java/app/tivi/util/TiviMvRxFragment.kt
```diff
@@ -0,0 +1,43 @@
...
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+        mvrxViewModelStore.saveViewModels(outState)
+    }
+
+    override fun onStart() {
+        super.onStart()
+        // This ensures that invalidate() is called for static screens that don't
+        // subscribe to a ViewModel.
+        postInvalidate()
```
elihart: this shouldn't be needed, mvrx does this automatically when the fragment starts

PR #214, opened 2018-09-03
Move over to AssistedInject

chrisbanes:
Currently not quite working due to https://github.com/square/AssistedInject/issues/47

### DIFF
--- a/app/src/main/java/app/tivi/showdetails/DetailsViewModelModule.kt
+++ b/app/src/main/java/app/tivi/showdetails/DetailsViewModelModule.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


app/src/main/java/app/tivi/showdetails/DetailsViewModelModule.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package app.tivi.showdetails
+
+import com.squareup.inject.assisted.dagger2.AssistedModule
+import dagger.Module
+
+@AssistedModule
+@Module(includes = [AssistedInject_DetailsViewModelModule::class])
```
hzsweers: This works? I've always gotten "cannot find symbol" errors when trying to reference generated code in an annotation o_O

vrendina: @hzsweers I see the same error with Dagger 2.17, if you drop back down to 2.16 it goes away. Not sure why yet. Tivi doesn't build if you upgrade to 2.17.

PR #237, opened 2018-11-05
Try out Google Cloud Build WIP

chrisbanes:
Many thanks to https://github.com/pixiteapps/android-cloud-build for most of the
implementation.

### DIFF
--- a/cloudbuild/cloudbuild.yaml
+++ b/cloudbuild/cloudbuild.yaml
@@ -1,0 +1,10 @@
+steps:
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/tar', './tar']
+
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/android', './android']
+
+images: 
+- 'gcr.io/$PROJECT_ID/tar'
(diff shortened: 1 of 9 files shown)


rharter:
I've updated the source repo to clean things up a little and also submitted my changes as a [PR](https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/127) to the cloud builders community repo.  Might be worth checking out the changes if you're planning to move forward with this.

cloudbuild/cloudbuild.yaml
```diff
@@ -0,0 +1,10 @@
+steps:
```
rharter: What's the reasoning for duplicating this in the repo?

It seems that the Cloud Build samples and docs mention storing this stuff in the repo, but that seems to conflict with our community's normal approach of using external dependencies instead of copying code, so I'd love to understand why, and if it's required.

I'd love to have something like a 1 button deploy (ala Heroku) for the docker images so developers can just deploy the image to their environment straight from the site and configure their yaml files accordingly.

chrisbanes: No idea, I was mainly just copying stuff in so that I can hack on things.

I made a few tweaks to the Android image, just to remove things I don't need (NDK, etc).

You could expose these all as separate repos, and then use hub.docker.com to build/host them. That way anyone can (theoretically) use them.

PR #238, opened 2018-11-08
Use $(...)  instead of legacy backticks for checksum.sh

philipphofmann:
Backtick command substitution `...` is legacy syntax with several issues.

- It has a series of undefined behaviors related to quoting in POSIX.
- It imposes a custom escaping mode with surprising results.
- It's exceptionally hard to nest.

$(...) command substitution has none of these problems, and is therefore strongly encouraged.

### DIFF
--- a/checksum.sh
+++ b/checksum.sh
@@ -7,7 +7,7 @@ fi
 touch $RESULT_FILE
 
 checksum_file() {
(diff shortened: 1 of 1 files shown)


chrisbanes:
Thanks!

PR #240, opened 2018-11-15
Use dexcount 'enabled' flag to run it only on CI

qwertyfinger:
New `enabled` [flag](https://github.com/KeepSafe/dexcount-gradle-plugin#configuration) was introduced in v0.8.4, which explicitly enables/disables counting, so it makes sense to use it for this purpose.
>enabled: When false, no build outputs will be counted. Defaults to true.


chrisbanes:
Thanks 👍

PR #241, opened 2018-11-15
Update Play Publisher and simplify its configuration

qwertyfinger:
`serviceAccountCredentials` configuration can be specified more succinctly.
Since there is no flavor-specific setup right now, I think it's reasonable to use the simpler way.


chrisbanes:
Thanks 👍

PR #271, opened 2019-02-22
Google Cloud Build implementation

rharter:
I've done a lot of work on the cloud builders since you started the integration, so here's the updated build script.

This is mostly functional, minus the deployment data since I don't have access to your keys.

Setup is much easier than before, simply: 

0. Setup `gcloud`, etc.
1. Install the cloud builders.

```bash
git clone git@github.com:pixiteapps/android-cloud-build /tmp/android-cloud-build
cd /tmp/android-cloud-build
gcloud builds submit .
```

2. Setup GCS buckets for cache, config and artifacts and update substitutions (via `--substitutions` works).

```yaml
  _ARTIFACT_BUCKET: black-circle-221-tivi-build-artifacts
  _CACHE_BUCKET: black-circle-221-tivi-build-cache
  _CONFIG_BUCKET: black-circle-221-tivi-build-config
```

### Note

This should still probably be considered experimental as it takes roughly 10-12 minutes to build, compared to the 4-5 minutes I'm seeing in your public CIrcleCI data.

I think this is due to the fact that GCB is running on an `n1-standard-1` instance, which only has a single vCPU and 3.75 GB of memory, as part of the free tier. 

Changing that to an `n1-highcpu-8` instance, which costs $0.016 / build-minute executes the build in 3-4 minutes. Based on the cost, and looking at how many builds you performed last month on CircleCI would cost $2.17 per month.  If you increased your build count by 60% to 50 builds per month, that would cost $3.20 per month.

While these costs aren't large, I'm going to suggest that GCB make at least the `n1-standard-2` VMs the free tier, since that would match up with the free offering of CircleCI (in CPU count) and Bitrise (in CPU count and Memory).

### DIFF
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -3,17 +3,28 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
+      <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
+      <module fileurl="file://$PROJECT_DIR$/base/base.iml" filepath="$PROJECT_DIR$/base/base.iml" />
       <module fileurl="file://$PROJECT_DIR$/base/base.iml" filepath="$PROJECT_DIR$/base/base.iml" />
       <module fileurl="file://$PROJECT_DIR$/base-android/base-android.iml" filepath="$PROJECT_DIR$/base-android/base-android.iml" />
       <module fileurl="file://$PROJECT_DIR$/buildSrc/buildSrc.iml" filepath="$PROJECT_DIR$/buildSrc/buildSrc.iml" />
       <module fileurl="file://$PROJECT_DIR$/data/data.iml" filepath="$PROJECT_DIR$/data/data.iml" />
+      <module fileurl="file://$PROJECT_DIR$/data/data.iml" filepath="$PROJECT_DIR$/data/data.iml" />
+      <module fileurl="file://$PROJECT_DIR$/data-android/data-android.iml" filepath="$PROJECT_DIR$/data-android/data-android.iml" />
       <module fileurl="file://$PROJECT_DIR$/data-android/data-android.iml" filepath="$PROJECT_DIR$/data-android/data-android.iml" />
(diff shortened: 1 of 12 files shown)


.idea/modules.xml
```diff
@@ -3,17 +3,28 @@
   <component name="ProjectModuleManager">
```
chrisbanes: Can you revert this file?

cloudbuild.yaml
```diff
@@ -0,0 +1,117 @@
...
+    # - 'JAVA_TOOL_OPTIONS="-Xmx4g"'
+    - 'BRANCH_NAME=$BRANCH_NAME'
+    - 'ARTIFACT_BUCKET=gs://${_ARTIFACT_BUCKET}'
+  waitFor:
+  - setup_env
+  - restore_cache
+
+- name: 'gcr.io/$PROJECT_ID/android:28'
+  id: build
+  args: ['./gradlew', 'bundleDebug', 'assembleDebug', '-Ptivi.versioncode=1']
```
chrisbanes: This should be BUILD_NUM (I'll manually hack it remain consistent with the CircleCI build numbers)

cloudbuild.yaml
```diff
@@ -0,0 +1,117 @@
...
+#
+# In order for this to run you need to deploy the Pixite Android cloud builders.
+#
+# ```bash
+# git clone git@github.com:pixiteapps/android-cloud-build /tmp/android-cloud-build
+# cd /tmp/android-cloud-build
+# gcloud builds submit .
+# ```
+
+steps:
```
chrisbanes: This is missing the deploy step, right?

.idea/modules.xml
rharter: Can do.

cloudbuild.yaml
rharter: Yeah, don't have keys and I've been using Cloud KMS so I didn't implement that (think I commented to that effect).

cloudbuild.yaml
rharter: Agreed.  The reason I hard coded this was because the current builder doesn't `eval` the parameters so doesn't support env variables in the args array.  I'm planning to update the builder to support that, since I haven't had the need for environment variables in args until I integrated your repo.

cloudbuild.yaml
chrisbanes: Ah yes sorry, missed the comment on the PR. I'll add

chrisbanes:
Committing this to a new `google-cloud-build` branch. I'll tweak the substitutions, etc

PR #279, opened 2019-03-14
Verify with the proper keystore for release build

jdtremblay:
I think the proper verification is to check if the `app-release.jks` file exists for release build.


jdtremblay:
I signed it!

chrisbanes:
Doh! Thanks!

PR #280, opened 2019-03-14
Fix OutOfMemoryError when building the first time

jdtremblay:
Gradle 5.0 change the default memory from 1g to 512m. This caused failed build with `OutOfMemoryError: GC overhead limit exceeded` error. Setting back the memory to 1g fix this issue. See https://github.com/gradle/gradle/issues/7746#issuecomment-439137327


chrisbanes:
Thanks!

PR #366, opened 2019-07-23
Fix checksum script

pavlospt:
Hey, I used the same script on an integration for CloudBuild and noticed that the `checksum.sh` was not working correctly. This PR fixes it 😃


chrisbanes:
Awesome, thanks!

PR #397, opened 2019-08-13
Use process scope for long-lived operations

### LINKED ISSUE #360: Why viewModelScope everywhere (for basically database actions)?
Hi, I want to ask about a design decision in this project.
All repository methods are transitively called viewModels via viewModelScope.launch { }

This however means when viewModel is destroyed (back press), then such action is canceled.

Why? I mean syncing, etc, all these actions seem to write to database, why is it a good idea to cancel / revert them?

What would it take for them not the get canceled on viewModelScope exit?

chrisbanes:
We now inject a CoroutineScope which is based off the ProcessLifecycleOwner lifecycle, meaning that the operations will not be cancelled prematurely. Flow/observable subscriptions still happen within the `viewModelScope`.

Closes #360

### DIFF
--- a/app/src/main/java/app/tivi/home/HomeActivityViewModel.kt
+++ b/app/src/main/java/app/tivi/home/HomeActivityViewModel.kt
@@ -24,11 +24,13 @@ import app.tivi.interactors.launchInteractor
 import app.tivi.trakt.TraktAuthState
 import app.tivi.trakt.TraktManager
 import app.tivi.TiviMvRxViewModel
+import app.tivi.inject.ProcessLifetime
 import app.tivi.interactors.launchObserve
 import com.airbnb.mvrx.MvRxViewModelFactory
 import com.airbnb.mvrx.ViewModelContext
 import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
@@ -40,7 +42,8 @@ class HomeActivityViewModel @AssistedInject constructor(
     @Assisted initialState: HomeActivityViewState,
     private val traktManager: TraktManager,
     private val updateUserDetails: UpdateUserDetails,
-    observeUserDetails: ObserveUserDetails
+    observeUserDetails: ObserveUserDetails,
+    @ProcessLifetime private val dataOperationScope: CoroutineScope
 ) : TiviMvRxViewModel<HomeActivityViewState>(initialState) {
     init {
         viewModelScope.launchObserve(observeUserDetails) {
@@ -55,7 +58,7 @@ class HomeActivityViewModel @AssistedInject constructor(
                     .distinctUntilChanged()
                     .onEach {
                         if (it == TraktAuthState.LOGGED_IN) {
-                            viewModelScope.launchInteractor(updateUserDetails,
+                            dataOperationScope.launchInteractor(updateUserDetails,
                                     UpdateUserDetails.Params("me", false))
                         }
                     }
--- a/app/src/main/java/app/tivi/inject/AppModule.kt
+++ b/app/src/main/java/app/tivi/inject/AppModule.kt
@@ -19,6 +19,8 @@ package app.tivi.inject
 import android.content.Context
 import android.content.SharedPreferences
 import android.os.Build
+import androidx.lifecycle.ProcessLifecycleOwner
+import androidx.lifecycle.coroutineScope
 import androidx.preference.PreferenceManager
 import app.tivi.BuildConfig
 import app.tivi.TiviApplication
@@ -26,6 +28,7 @@ import app.tivi.util.AppCoroutineDispatchers
 import dagger.Module
 import dagger.Provides
 import io.reactivex.disposables.CompositeDisposable
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import org.threeten.bp.ZoneId
 import org.threeten.bp.format.DateTimeFormatter
@@ -123,5 +126,11 @@ class AppModule {
         return DateTimeFormatter.ofPattern(dateF.toPattern())
                 .withLocale(Locale.getDefault())
                 .withZone(ZoneId.systemDefault())
+    }
+
+    @Provides
+    @ProcessLifetime
+    fun provideLongLifetimeScope(): CoroutineScope {
+        return ProcessLifecycleOwner.get().lifecycle.coroutineScope
     }
 }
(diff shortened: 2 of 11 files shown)


ursusursus:
Isnt this a leaking? (Launching a coroutine on dataScope in viewModel with viewModel local stuff in the lambda)

chrisbanes:
Yeah you're right, due to the `ObservableLoadingCounter` object. Everything else should be fine.

ursusursus:
Well, using join should be fine, and also the way you write interactors is also non leaking (since they dont have a reference to viewModel), however, not sure about the design, as, injecting dataScope into viewModel, might lend it self to people using dataScope.launch { ... } there, hence leaking this (viewModel)

Why not have the upper (domain) layers decide its operations scope? (i.e. let the scope be a impl. detail of such Repository etc)

chrisbanes:
@ursusursus totally agree, this was a quick fix but I want to move this to the repository layer. This is also related to #396

ursusursus:
If I understand that correctly, youre trying to share the network request between callsites?
Also, one more issue I think youre going to encounter is "reconnecting" with such actions after say going back and then forward to the same UI (different instance). Since now action is still in flight thanks to dataScope, but the new UI instance has no idea of this. (Unless reconnected via code in that PR, but that I feel is going to be awkward, since that has to be done in ctor). How about sideffecting to a relay / channel, and lets the lower level only observer that? Same as views observe replayed observables exposed by viewmodel

chrisbanes:
Are you describing moving to using the actor pattern? (which is an option I'm thinking about).

ursusursus:
Well, Im trying to design this whole thing in my apps also, so im looking for validation / giving you ideas. I came to recognize 2 issues, 1st is to have the Repository or any other domain object own its own scope. The classical Rx solution would be to not expose a cold observable, but have a imperative "trigger" method (subscribing a cold observable to `this` + returning Unit), where results would be observer indirectly via some other hot observable (so triggerFoo() + fooStateObservable pairs). This works for consuming layers coming and going, since fooStateObservable is replayed. However the Unit trigger method sucks because you just lost comosability (you cannot flatmap over this). Here coroutines Deffered works well. since you get that disconnecability (not being canceled when client goes away, + you still keep composability of suspend function) My prototype looks like this

```
class Repository {
    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)

    suspend fun doSomething(): String = scoped x@{
        foo()
        return@x "Something"
    }


    private suspend fun <T> scoped(body: suspend CoroutineScope.() -> T): T {
        val deferred = scope.async(block = body)
        return deferred.await()
    }

    private fun foo() {
        Log.d("Default", "foo")
    }

    fun close() {
        scope.cancel()
    }
}

```

However there is still the issue of how to let new client instance know about the state of doSomething, so my solution for now is simply to sideffect

```
suspend fun doSomething(): String = scoped x@{
        doSomethingStateChannel.send(SomethingState.STARTED)
        foo()
        doSomethingStateChannel.send(SomethingState.IDLE)
        return@x "Something"
    }
```
(Or maybe your reconnecting to the inflight doSomething would be better, however im not sure about the api, since you need to call that function twice, and you might actually trigger it when you only want to reconnect etc)

Let me know what you think 

(Im not familiar with actor pattern yet)

chrisbanes:
I thought about something like that, but it explicitly goes against what Roman talks about here: https://medium.com/@elizarov/coroutine-context-and-scope-c8b255d59055. Specifically: 

> Do not do this! It makes the scope in which the coroutine is launched opaque and implicit, capturing some outer Job to launch a new coroutine without explicitly announcing it in the function signature. A coroutine is a piece of work that is concurrent with the rest of your code and its launch has to be explicit.

I think the place to move the scoping is within the use-cases/interactors. They are then responsible for moving the actual data operation out of the `viewModelScope`

ursusursus:
Yes im aware of that and I dont understand his reasoning. Maybe its somehow equivalent to singletons vs DI scopes? No sure.

I still think its implementation detail of a given class. To me, its the same as there is viewModelScope in viewModel etc, precedence is there already. And it makes all layers behave the same, each has its own scope, and wider lifecycle than the consuming layer etc. View layer is the same just happens to be the terminal one

chrisbanes:
Do you want the layers to behave the same though? Personally I want to keep the data layers as clean as possible, and moving the scoping into them makes them super cluttered.

...but I do want control over what calls through to my data layer (in my case, interactors).

I think what I currently have is OK, in terms of the relationship between `VM` <> `Interactor` <> `Repository`. Where it currently fails is observing the state of the interactor, which should be as simple as exposing a `Flow<Status>` from the interactor, and observing that in the VM (using the `viewModelScope`

``` kotlin
class Interactor @Inject constructor(val scope: CoroutineScope) {
    private val statusChannel = ConflatedBroadcastChannel<Status>()
    val status = statusChannel.asFlow()
    
    operator fun invoke(params: P) {
        scope.launch {
            status.send(Status.STARTED)
            // Call data layer, etc
            status.send(Status.FINISHED)
        }
    }
}
```

``` kotlin
class BlahViewModel(val interactor: Interactor) {
    init {
        viewModelScope.launch {
            interactor.status.collect { // show loading or something }
        }
    }

    fun onClick() {
        interactor()
    }
}
```

ursusursus:
Im not super sure, but in my mind the best architecture would be one that scales infinitely, i.e. one where components can be broken apart recursively. Similar to code package structure - package too complex? split it into subpackages etc. I think Uber RIBs does something like this., or the lego's from square.

So, maybe this is somewhat clashing with Clean, since to me honestly Interactors feel like decomposition of big ViewModels, not the actual domain layer, since theyre lifecycle is the same as viewmodel, hence this scoping issue, since if their lifecycle was wider then youd just scope such actions to that

But I dont know

chrisbanes:
I think of interactors the other way, they combine your data layer to achieve something. Some of my interactors simply call one method ([example](https://github.com/chrisbanes/tivi/blob/master/interactors/src/main/java/app/tivi/interactors/UpdateEpisodeDetails.kt)), others are more complex and combine multiple repositories [example](https://github.com/chrisbanes/tivi/blob/master/interactors/src/main/java/app/tivi/interactors/UpdateFollowedShows.kt).

> So, maybe this is somewhat clashing with Clean, since to me honestly Interactors feel like decomposition of big ViewModels, not the actual domain layer, since theyre lifecycle is the same as viewmodel, hence this scoping issue, since if their lifecycle was wider then youd just scope such actions to that

I assume you're talking specifically about how I use them? If so, I totally agree which is why my proposal [above](https://github.com/chrisbanes/tivi/pull/397#issuecomment-521089544) breaks that coupling apart (i.e. the `invoke()` is not a suspending fun)

ursusursus:
Yes I like that sample, but who clears that scope param or rather what is that scope?

chrisbanes:
I'm just implementing this now, I'll upload a PR once I'm done :+1:

PR #506, opened 2019-11-04
Move to GitHub Actions

chrisbanes:


PatrickDattilio:
You can replace the last step with `run: ./gradlew bundleDebug assembleDebug` instead of using the `setup android` action and save yourself the several minute docker download. `ubuntu-latest` already has gradle installed.

chrisbanes:
Gradle being installed isn't that important, it's having the Android SDK available. I haven't looked but I doubt the standard VM has it.

PatrickDattilio:
Sorry, meant to say both Android SDK (10-29) and Gradle are installed on all the VMs. https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners

chrisbanes:
Ahhh, interesting! Thanks for letting me know.

chrisbanes:
Unfortunately GitHub Actions is too slow to be usable at the moment.

The `cache` action only supports caches of up to 400MB currently, which is less than what we need (~500MB), resulting in each run being completed without a Gradle build/jar `cache`. Tracking bug: https://github.com/actions/cache/issues/6

Also, there is currently no way to store complex workspaces across jobs. We can save a single directory using `upload-artifact`, but that is not useful for Gradle builds, which use separate `build` directories in each module. Tracking bug: https://github.com/actions/upload-artifact/issues/3

PR #526, opened 2019-11-13
Update to Coil v0.9.x

chrisbanes:
DO NOT MERGE until 0.9.0 is out.


common-imageloading/src/main/java/app/tivi/common/imageloading/SaturatingTransformation.kt
```diff
@@ -0,0 +1,64 @@
...
+                animator.doOnEnd {
+                    continuation.resume(Unit) { animator.cancel() }
+                }
+                animator.start()
+
+                continuation.invokeOnCancellation { animator.cancel() }
+                adapter.drawable = drawable
+            }
+        } else {
+            adapter.drawable = drawable
```
colinrtwhite: `drawable` is only null in if the request fails and its `error` drawable is null (I probably should add this to the docs). 

If you want to skip the transition in that case you could return `null` from `Transition.Factory.newTransition` if `event == Transition.Event.ERROR`.

PR #528, opened 2019-11-13
Increase java heap size to 2 GB

LivingWithHippos:
Avoid "GC overhead limit exceeded" error while building. If this is not specified it should default to 1 GB, I wasn't able to run tivi without this.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,6 +14,7 @@
 org.gradle.configureondemand=true
 org.gradle.caching=true
 org.gradle.parallel=true
+org.gradle.jvmargs=-Xmx2g
 
 # AndroidX
 android.useAndroidX=true


LivingWithHippos:
@googlebot I signed it!

chrisbanes:
Thanks! I intentionally didn't set this due to the priority order of Gradle handling the `jvmargs` properties.

Currently, our CI script overrides the max heap to ~5GB using `JAVA_TOOL_OPTIONS`, but `org.gradle.jvmargs` in `gradle.properties` has a higher priority so the env variable is ignored (the same thing happens when using `GRADLE_OPTS`).

I'll merge this and see what can be done for CI.

LivingWithHippos:
According to https://docs.gradle.org/current/userguide/build_environment.html command line parameters have higher priority than gradle.properties.
At https://docs.gradle.org/current/userguide/command_line_interface.html#environment_options there's the parameter
```
-Dorg.gradle.jvmargs

    Set JVM arguments.
```
maybe this can be useful ;P

chrisbanes:
Ah yes, thanks for researching.

I went a different way, which I think works better with version control: https://github.com/chrisbanes/tivi/commit/a0271ec55f2f77d3a991590c5393e197ad692922

This allows me to extract the config to a file, which is easier to update separately.

PR #544, opened 2019-12-15
Converting episode details to use Jetpack Compose

chrisbanes:


ui-episodedetails-compose/src/main/java/app/tivi/episodedetails/EpisodeDetails.kt
```diff
@@ -0,0 +1,251 @@
...
+import app.tivi.data.entities.PendingAction
+import app.tivi.data.entities.Season
+import app.tivi.episodedetails.compose.R
+import app.tivi.util.TiviDateFormatter
+
+/**
+ * This is a bit of hack. I can't make `ui-episodedetails` depend on any of the compose libraries,
+ * so I wrap `setContext` as my own function, which `ui-episodedetails` can use.
+ */
+fun composeEpisodeDetails(
```
clarabayarri: Shouldn't this be @Composable? You are emitting Composables below, I'm surprised this compiles/works

chrisbanes: Not sure? This is just a wrapper around `setContent()` which I assume doesn’t need to be `@Composable`

chrisbanes:
Ignore the build failures, they due to issues in AGP 4.0.0 [#1](https://issuetracker.google.com/issues/148005697), [#2](https://issuetracker.google.com/issues/148172219)

build.gradle
```diff
@@ -61,6 +61,8 @@ allprojects {
         google()
         mavenCentral()
         jcenter()
+        maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
```
manuelvicnt: I don't think you need this anymore as the Compose-specific content in kotlin-stdlib has landed in 1.3.61

common-ui-compose/src/main/java/app/tivi/common/compose/gradientscrim.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.ui.core.Draw
+import androidx.ui.geometry.Offset
+import androidx.ui.graphics.Color
+import androidx.ui.graphics.LinearGradientShader
+import androidx.ui.graphics.Paint
+import androidx.ui.unit.toRect
+import kotlin.math.pow
+
+@Composable
+fun GradientScrim(baseColor: Color, numStops: Int = 16) {
```
manuelvicnt: Is there a special reason of why the file names of the Compose module are lowercase?

chrisbanes: Not really, just a convention I've picked up. I quite like it since it's then obvious what files contain lots of top-level functions/random utils, vs a proper `class`

build.gradle
chrisbanes: Good catch, I'll remove.

common-ui-compose/src/main/java/app/tivi/common/compose/gradientscrim.kt
manuelvicnt: Cool! Couldn't resolve this comment :man_shrugging:

PR #550, opened 2020-01-08
Update build.yml

imbsky:
Since checkout action v2, the depth is 1 by default.


imbsky:
@googlebot I signed it!

imbsky:
@googlebot I signed it!

chrisbanes:
Thanks!

PR #555, opened 2020-02-04
wondering why you used a .txt file instead of a variable

igor-resende-ifood:
Nice job, by the way! Excellent approach.. =)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,10 @@ jobs:
         env:
           ENCRYPT_KEY: ${{ secrets.ENCRYPT_KEY }}
 
-      - name: Generate cache key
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -25,7 +25,10 @@ jobs:
         env:
           ENCRYPT_KEY: ${{ secrets.ENCRYPT_KEY }}
 
-      - name: Generate cache key
+# Hello Chris, I'm opening this PR just to ask why are you using a .txt file instead of using a variable to store the hashes..
```
JoseAlcerreca: nit: padding.

chrisbanes:
If you found that a variable is quicker (I'd guess so too), please send a proper PR :smile:

PR #556, opened 2020-02-10
Migrate to using Store

chrisbanes:
Super early days. Currently on the 'show' and 'show images' entities are migrated over.

Next up is the show lists:

- [x] Trending shows (paginated)
- [x] Popular shows (paginated)
- [x] Watched shows (paginated)
- [x] Recommended shows
- [ ] Followed shows
- [x] Related shows


chrisbanes:
FYI: @yigit and @digitalbuddha

data/src/main/java/app/tivi/data/StoreExtensions.kt
```diff
@@ -0,0 +1,66 @@
...
+    forceFresh: Boolean = false,
+    crossinline doFreshIf: (Output) -> Boolean = { false }
+): Output {
+    return if (forceFresh) {
+        // If we're forcing a fresh fetch, do it now
+        fresh(key)
+    } else {
+        // Else we'll check the current cached value
+        val cached = cachedOnly(key)
+        if (cached == null || doFreshIf.invoke(cached)) {
```
yigit: nit, feel free to ignore.
`doFreshIf(cached)`

yigit: @digitalbuddha it might make sense for us to provide an APi that'll validate cache values for easy invalidation.

data/src/main/java/app/tivi/data/StoreExtensions.kt
```diff
@@ -0,0 +1,66 @@
...
+package app.tivi.data
+
+import com.dropbox.android.external.store4.Store
+import com.dropbox.android.external.store4.StoreRequest
+import com.dropbox.android.external.store4.StoreResponse
+import com.dropbox.android.external.store4.fresh
+import kotlinx.coroutines.flow.filterNot
+import kotlinx.coroutines.flow.first
+
+suspend inline fun <Key : Any, Output : Any> Store<Key, Output>.fetch(
```
yigit: alternatively, you can write this as a reactive stream:

```
val request = if (forceRefresh) {
    StoreRequest.fresh(key)
} else {
    StoreRequest.cached(key, refresh = true)
}
return store.stream(request).filter { result ->
    when(result) {
       is StoreResponse.Loading -> false // drop
       is StoreResponse.Data -> {
            if(result.origin == ResponseOrigin.Fetcher) {
                     true
            } else if (doFreshIf(result.requireData()) {
                    false
            } else {
                   true
            }
       }
       else -> true // let errors pass
    }
}.first().requireData()
```

data/src/main/java/app/tivi/data/daos/EntityDao.kt
```diff
@@ -21,19 +21,34 @@ import androidx.room.Insert
...
 
     @Update
-    suspend fun update(entity: E)
+    abstract suspend fun update(entity: E)
 
     @Delete
-    suspend fun delete(entity: E): Int
+    abstract suspend fun deleteEntity(entity: E): Int
+
+    suspend fun insertOrUpdate(entity: E): Long {
```
yigit: i think this is fixed in latest room such that you can have a concrete implementation in an interface. might be the upcoming patch release though.

data/src/main/java/app/tivi/data/repositories/shows/ShowsModule.kt
```diff
@@ -16,18 +16,92 @@
...
+        showDao: TiviShowDao,
+        @Trakt traktShowDataSource: ShowDataSource,
+        @Tmdb tmdbShowDataSource: ShowDataSource
+    ): ShowStore {
+        return StoreBuilder.fromNonFlow { showId: Long ->
+            val localShow = showDao.getShowWithId(showId)
+                ?: throw IllegalArgumentException("No show with id $showId in database")
+
+            coroutineScope {
+                val traktResult = async {
```
yigit: you "probably" want to do this merging when writing it to disk. i'm not well aware of this code but if you are just trying to update local value carefully (assuming if tmdbShowDataSource is just local), then it is better to do it before it is inserted to database.

Store considers the persister as source of truth so it will never return values from the fetcher as is.

data/src/main/java/app/tivi/data/daos/EntityDao.kt
chrisbanes: Cool, I'm on the latest but will migrate when it's out

PR #560, opened 2020-02-24
New 'save data' preference

chrisbanes:


common-imageloading/src/main/java/app/tivi/common/imageloading/TmdbImageEntityCoilMapper.kt
```diff
@@ -37,7 +38,7 @@ class TmdbImageEntityCoilMapper @Inject constructor(
 
     override fun map(data: TmdbImageEntity, size: Size): HttpUrl {
         val width = if (size is PixelSize) {
-            if (powerController.canFetchHighResolutionImages()) {
+            if (powerController.shouldSaveData() is SaveData.Enabled) {
```
MartelliEnrico: The logic was inverted here!

PR #572, opened 2020-03-26
Updated the development setup instructions.

Sathawale27:
As not using fabric anymore.

**Reference revision:**
Remove Fabric and move to Firebase client lib
Revision - 1f5f4cf9cc0851335e942efb0279d57c0c6ee770


chrisbanes:
Thanks!

PR #582, opened 2020-03-28
Update to Compose snapshot #6342451

chrisbanes:


common-ui-compose/src/main/java/app/tivi/common/compose/imageloading.kt
```diff
@@ -120,12 +121,14 @@ fun LoadNetworkImageWithCrossfade(
...
                 // one every time
                 val cf = ColorMatrixColorFilter(matrix)
-                Box(modifier = modifier + AndroidColorMatrixImagePainter(image, cf).toModifier())
+                Box(
+                    modifier = modifier.onPositioned { childSize = it.size }
+                        .paint(AndroidColorMatrixImagePainter(image, cf))
+                )
             } else {
-                Box(modifier = modifier)
+                Box(modifier = modifier.onPositioned { childSize = it.size })
```
andkulikov: could be nicer to do something like
var modifier = modifier.onPositioned { childSize = it.size }
if (image) {
    ...
   modifier.paint(AndroidColorMatrixImagePainter(image, cf))
}
Box(modifier = modifier)

everytime you have if else blocks emitting different composables the previous Box should be completely disposed and the new Box recreated where in fact you can just update a modifier for the same Box

chrisbanes: Thanks, done!

PR #583, opened 2020-04-01
Update show details to Compose

chrisbanes:


chrisbanes:
Ready for review. Demo video: https://youtu.be/t-F4Q052QoY

common-ui-compose/src/main/java/app/tivi/common/compose/ExpandSummary.kt
```diff
@@ -0,0 +1,64 @@
...
+    text: String,
+    textStyle: TextStyle = MaterialTheme.typography.body2,
+    collapsedMaxLines: Int = 4,
+    expandedMaxLines: Int = Int.MAX_VALUE,
+    modifier: Modifier = Modifier
+) {
+    var canExpand by stateFor(text) { true }
+
+    Box(modifier = Modifier.ripple(bounded = true, enabled = canExpand)) {
+        var expanded by state { false }
```
objcode, line 46: Suggest hoisting this as it seems reasonable callers may want to control the state.

chrisbanes: Good call, done 👍

PR #587, opened 2020-04-19
Removed unused constraints

Sathawale27:
Removed unused constraints.

### DIFF
--- a/app/src/main/res/layout/activity_home.xml
+++ b/app/src/main/res/layout/activity_home.xml
@@ -40,9 +40,6 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             app:labelVisibilityMode="labeled"
-            app:layout_constraintBottom_toBottomOf="parent"
-            app:layout_constraintLeft_toLeftOf="parent"
-            app:layout_constraintRight_toRightOf="parent"
             app:menu="@menu/home_bottomnav" />
 
     </LinearLayout>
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/activity_home.xml
```diff
@@ -40,9 +40,6 @@
             android:layout_width="match_parent"
```
chrisbanes, line 40: It's better to keep the explicit constraints and set the `layout_width` to `match_constraint` or `0dp` (same thing)

ui-showdetails/src/main/res/layout/view_holder_details_related_item.xml
```diff
@@ -60,10 +60,6 @@
             android:padding="@dimen/spacing_small"
             android:text="@{tiviShow.title}"
             android:textAppearance="?attr/textAppearanceBody2"
-            app:layout_constraintBottom_toBottomOf="parent"
```
chrisbanes, line 63: It's better to keep the explicit constraints and set the `layout_width` and `layout_height` to `match_constraint` or `0dp` (same thing)

app/src/main/res/layout/activity_home.xml
Sathawale27: Yeah, Right! I will do that. So, do you want me to update the layouts with `constraint layout`?

chrisbanes: Ohhhh, I forgot that I converted these away from ConstraintLayout! 🤦

Sathawale27: So, do you want to convert those again to ConstraintLayout? Was there any particular reason to convert away from ConstraintLayout?

chrisbanes: No particular reason, other than simplicity.

Sathawale27: Okay.

chrisbanes:
Closing these, since we're focused on migrating Tivi to Compose

PR #591, opened 2020-05-01
Update to Compose 0.1.0-dev10

chrisbanes:
Had to pin our Kotlin source level to 1.3 since Compose now depends on the 1.4 Kotlin compiler


florina-muntenescu:
LGTM

PR #617, opened 2020-05-19
Update to Compose v0.1.0-dev11

chrisbanes:


common-ui-compose/src/main/java/app/tivi/common/compose/gradientscrim.kt
```diff
@@ -19,38 +19,44 @@ package app.tivi.common.compose
...
-        paint.shader = LinearGradientShader(
-            Offset.zero,
-            Offset(0f, size.height.value),
-            colors
-        )
+    drawWithContent {
+        drawContent()
 
-        drawRect(size.toRect(), paint)
+        drawCanvas { canvas, pxSize ->
```
nickbutcher, line 53: This shouldn't need `drawCanvas` can use a `VerticalGradient` brush and save the `Paint` allocation. See Owl.

PR #618, opened 2020-05-19
Some more tweaks enabled in dev11

chrisbanes:


common-ui-compose/src/main/java/app/tivi/common/compose/AutoSizedCircularProgressIndicator.kt
```diff
@@ -0,0 +1,85 @@
...
+
+        CircularProgressIndicator(
+            strokeWidth = (diameter * strokePerDiameterDp).coerceAtLeast(1.dp),
+            color = color
+        )
+    }
+}
+
+// Default stroke size
+private val defaultStrokeWidth = 4.dp
```
nickbutcher, line 51: I've been using PascalCase for constants like the material libs.

common-ui-compose/src/main/java/app/tivi/common/compose/AutoSizedCircularProgressIndicator.kt
```diff
@@ -0,0 +1,85 @@
...
+import androidx.ui.unit.min
+
+@Composable
+fun AutoSizedCircularProgressIndicator(
+    modifier: Modifier = Modifier,
+    color: Color = MaterialTheme.colors.primary
+) {
+    WithConstraints(modifier) {
+        val diameter = with(DensityAmbient.current) {
+            min(constraints.maxWidth, constraints.maxHeight).toDp() - defaultPadding
```
nickbutcher, line 40: Hoist the padding?

common-ui-compose/src/main/java/app/tivi/common/compose/AutoSizedCircularProgressIndicator.kt
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// Default stroke size
+private val defaultStrokeWidth = 4.dp
+// Preferred diameter for CircularProgressIndicator
+private val defaultDiameter = 40.dp
+// Default padding added by CircularProgressIndicator
+private val defaultPadding = 4.dp
+
+private val strokePerDiameterDp = defaultStrokeWidth / defaultDiameter
```
nickbutcher, line 57: Don't love the name. What about `strokeFraction`?

common-ui-compose/src/main/java/app/tivi/common/compose/AutoSizedCircularProgressIndicator.kt
chrisbanes: The padding is an internal constant in `CircularProgressIndicator` which I need to take into account to find the 'true' diameter.. Not much point hoisting it as it's not really a parameter imo.

common-ui-compose/src/main/java/app/tivi/common/compose/gradientscrim.kt
```diff
@@ -17,46 +17,56 @@
...
             val x = i * 1f / (numStops - 1)
-            val opacity = x.pow(3.0f)
+            val opacity = x.pow(decay)
             color.copy(alpha = baseAlpha * opacity)
         }
     }
 
+    var height by state { 0f }
+
+    val shader = remember(height) {
```
nickbutcher, line 59: Also based on `colors`
```suggestion
    val shader = remember(colors, height) {
```

PR #627, opened 2020-06-08
Render ER diagrams with FloorPlan

julioz:
[FloorPlan](https://github.com/julioz/FloorPlan) translates Room schemas into entity-relationship diagrams, to make it easier to inspect changes in the persistance-layer schema of the app.

Here, I am configuring the first version of FloorPlan to output `SVG` files by mimicking the Room directory structure, into a sibling directory under the `:data-android` module. 
The first commit is the plugin configuration, the second is checking in the actual diagrams.

You can manually generate the diagrams by running

```
./gradlew :data-android:generateFloorPlan
```

Optionally, we could also make `compileDebugJava`/`kaptDebugKotlin` depend on `generateFloorPlan`, and we would get automatic diagram generation on builds for that module, similarly to how [Room hooks into the build process to generate the JSON schemas](https://stackoverflow.com/a/57103344/12511149).
Let me know your thoughts on this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,6 +47,8 @@ buildscript {
         classpath Libs.AndroidX.Navigation.safeArgs
 
         classpath "com.github.triplet.gradle:play-publisher:2.8.0-SNAPSHOT"
+
+        classpath Libs.FloorPlan.plugin
     }
 }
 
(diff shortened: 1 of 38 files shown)


chrisbanes:
Hi Julio, FloorPlan looks like a really nice tool! 👏. Unfortunately I don't have a need to visualise the database schema in Tivi so I won't be merging this in (the build is already slow enough).

PR #639, opened 2020-06-16
Trying out new Insets implementation

chrisbanes:


common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,139 @@
...
+import androidx.ui.unit.Density
+import androidx.ui.unit.Dp
+import androidx.ui.unit.dp
+
+/**
+ * Main holder of our inset values.
+ *
+ * TODO add other inset types (IME, visibility, etc)
+ */
+class DisplayInsets {
```
nickbutcher, line 46: Rather than having this holder class, what about just making each property of `Insets` a `mutableStateOf` and then updating each?

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,139 @@
...
+    val density = DensityAmbient.current
+
+    val displayInsets = remember { DisplayInsets() }
+
+    onCommit(view.id) {
+        ViewCompat.setOnApplyWindowInsetsListener(view) { _, insets ->
+            displayInsets.updateFrom(insets, density)
+            // Return the unconsumed insets
+            insets
+        }
```
nickbutcher, line 76: Doesn't this need need `requestApplyInsets` here?

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,139 @@
...
+) {
+    companion object {
+        val Zero = Insets()
+    }
+}
+
+val InsetsAmbient = staticAmbientOf { DisplayInsets() }
+
+@Composable
+fun ProvideInsets(children: @Composable () -> Unit) {
```
nickbutcher, line 65: nit: think we're moving away from `children` to `content`

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Yep, added 👍

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Good idea, done.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Done 👍

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,300 @@
...
+import androidx.ui.layout.width
+import androidx.ui.unit.Density
+import androidx.ui.unit.dp
+
+/**
+ * Main holder of our inset values.
+ */
+@Stable
+class DisplayInsets {
+    val systemBars by lazy(LazyThreadSafetyMode.NONE) { Insets() }
```
adamp, line 48: Why use lazy here?

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,300 @@
...
+/**
+ * Allows conditional setting of [insets] on each dimension.
+ */
+private inline fun Modifier.insetsPadding(
+    insets: Insets,
+    left: Boolean = false,
+    top: Boolean = false,
+    right: Boolean = false,
+    bottom: Boolean = false
+) = absolutePadding(
```
adamp, line 177: Write a private pixel-based padding modifier so that you don't have to convert back and forth from dp. Stay in pixels from the system's reported insets all the way through.

This will also let you put the insets state read in layout rather than in composition when this modifier is constructed.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,300 @@
...
+            windowInsets
+        }
+        view.doOnAttach { it.requestApplyInsets() }
+
+        onDispose {
+            ViewCompat.setOnApplyWindowInsetsListener(view, null)
+        }
+    }
+
+    Providers(InsetsAmbient provides displayInsets) {
```
adamp, line 113: Consider factoring out a producer function that just returns the remembered and managed DisplayInsets, so that this abstraction doesn't have to be responsible for managing a Providers call and content function.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,300 @@
...
+    val displayInsets = remember { DisplayInsets() }
+
+    onCommit(view) {
+        ViewCompat.setOnApplyWindowInsetsListener(view) { _, windowInsets ->
+            displayInsets.systemBars.updateFrom(windowInsets.systemWindowInsets, density)
+
+            // Return the unconsumed insets
+            windowInsets
+        }
+        view.doOnAttach { it.requestApplyInsets() }
```
adamp, line 106: Depending on how we decide to manage detached compose views in the future this might not be sufficient. Consider a full/long-lived OnAttachStateChangedListener.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,300 @@
...
+
+        onDispose {
+            ViewCompat.setOnApplyWindowInsetsListener(view, null)
```
adamp, line 109: Consider checking that the current listener is still yours before setting it to null here

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,300 @@
...
+    var bottom by mutableStateOf(0.dp, StructurallyEqual)
+        internal set
+
+    /**
+     * TODO: doesn't currently work
+     */
+    var visible by mutableStateOf(true)
+        internal set
+
+    override fun equals(other: Any?): Boolean {
```
adamp, line 68: Are the equals/hashCode implementations still needed or are these leftover from a version where this was immutable?

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,300 @@
...
+    right = if (right) insets.right else 0.dp,
+    bottom = if (bottom) insets.bottom else 0.dp
+)
+
+/**
+ * Updates our mutable state backed [Insets] from an Android system insets.
+ */
+private fun Insets.updateFrom(
+    insets: androidx.core.graphics.Insets,
+    density: Density
```
adamp, line 189: See other comments re. staying in pixels instead of converting

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,300 @@
...
+ * layout constraints.
+ *
+ * @param side The navigation bar side to use as the source for the width.
+ */
+fun Modifier.navigationBarWidth(side: HorizontalSide) = composed {
+    // TODO: Move to Android 11 WindowInsets APIs when they land in AndroidX.
+    // It currently assumes that nav bar == left/right
+    val insets = InsetsAmbient.current
+    when (side) {
+        HorizontalSide.Left -> Modifier.width(insets.systemBars.left)
```
adamp, line 297: See other comments about staying in pixels and deferring the state reads to layout

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: There's no getter :sadface:

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Nah, I can remove them

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: No particular reason, removed.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Removed this call, since it doesn't do much anyway.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Done.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: I tried, but I'm not sure this is going to work as-is. A view can only have one `OnApplyWindowInsetsListener` set, so if the producer function (say called `displayInsets()`) is called in multiple places, only one them will actually work.

I could shove a multiplexing `OnApplyWindowInsetsListener` or a `InsetsController` to an Ambient, and then have each `displayInsets()` subscribe to it. WDYT?

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Done in the latest commit 👍

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Done.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,415 @@
...
+        }
+        view.addOnAttachStateChangeListener(attachListener)
+
+        if (view.isAttachedToWindow) {
+            // If the view is already attached, we can request an inset pass now
+            view.requestApplyInsets()
+        }
+
+        onDispose {
+            view.removeOnAttachStateChangeListener(attachListener)
```
nickbutcher, line 106: Remove the insets listener here?

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,415 @@
...
+        val vertical = top + bottom
+
+        val placeable = measurable.measure(constraints.offset(-horizontal, -vertical))
+
+        val width = (placeable.width + horizontal)
+            .coerceIn(constraints.minWidth, constraints.maxWidth)
+        val height = (placeable.height + vertical)
+            .coerceIn(constraints.minHeight, constraints.maxHeight)
+        return layout(width, height) {
+            placeable.placeAbsolute(left, top)
```
nickbutcher, line 312: How does this handle RTL?

chrisbanes: Insets are always absolute, so no need.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
```diff
@@ -0,0 +1,415 @@
...
+            val resolvedMaxHeight = if (heightSide != null) {
+                targetConstraints.maxHeight
+            } else {
+                min(constraints.maxHeight, targetConstraints.minHeight)
+            }
+            Constraints(resolvedMinWidth, resolvedMaxWidth, resolvedMinHeight, resolvedMaxHeight)
+        }
+        val placeable = measurable.measure(wrappedConstraints)
+        return layout(placeable.width, placeable.height) {
+            placeable.place(0, 0)
```
nickbutcher, line 376: Using auto-mirrored `place` here.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: I had this originally, but I'm not sure it's right because technically the dev could have set their own listener.

Ideally this would live in `ComposeView` itself (or whatever it is called), and then it forwards as appropriate from `onApplyWindowInsets`.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
nickbutcher: Do we need the `DisplayInsets` class? Can't we just hold an `Insets` instance in the ambient?

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Good catch, fixed.

common-ui-compose/src/main/java/app/tivi/common/compose/Insets.kt
chrisbanes: Yep, eventually I'll add `IME` insets. I'll add the gesture insets now.

nickbutcher:
Is this ready for review or still WIP?

chrisbanes:
Yep, ready for review.

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
```diff
@@ -14,64 +14,410 @@
...
+ * This will automatically listen for [WindowInsetsCompat] on the host view.
+ *
+ * @param consumeInsets Whether to consume the insets which are dispatched to the host view.
+ */
 @Composable
-fun ProvideInsets(
-    liveData: LiveData<WindowInsetsCompat?>,
-    children: @Composable () -> Unit
+fun ProvideDisplayInsets(
+    consumeInsets: Boolean = true,
```
nickbutcher, line 84: Would you ever want to consume one type of insets but not another?

chrisbanes: Not from API 30 (and soon androidx.core). The `consumeXXXX()` functions have been deprecated, and apps should now return `WindowInsets.CONSUMED` if they want to mark the insets as consumed.

I'll update this impl once the core update has gone out, which should be 12th August.

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
```diff
@@ -14,64 +14,410 @@
...
+
+/**
+ * Selectively apply additional space which matches the width/height of any system bars present
+ * on the respective edges of the screen.
+ *
+ * @param enabled Whether to apply padding using the system bar dimensions on the respective edges.
+ * Defaults to `true`.
+ */
+fun Modifier.systemBarsPadding(enabled: Boolean = true) = composed {
+    insetsPadding(
```
nickbutcher, line 137: What about skipping `insetsPadding` if enables is false?

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
```diff
@@ -14,64 +14,410 @@
...
+ * of the content.
+ */
+fun Modifier.statusBarPadding() = composed {
+    insetsPadding(insets = DisplayInsetsAmbient.current.systemBars, top = true)
+}
+
+/**
+ * Apply additional space which matches the height of the navigation bar height
+ * along the [bottom] edge of the content, and additional space which matches the width of
+ * the navigation bar on the respective [left] and [right] edges.
```
nickbutcher, line 157: “width of the nav bar” sounds misleading as it's the overlap with insets right?

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
```diff
@@ -14,64 +14,410 @@
...
+        }
+    }
+
+    Providers(DisplayInsetsAmbient provides displayInsets) {
+        content()
+    }
+}
+
+/**
+ * Selectively apply additional space which matches the width/height of any system bars present
```
nickbutcher, line 130: What about replacing 'additional space' with 'padding' throughout. Right now comments mix these terms. With `Modifier.padding` being the lib's chosen naming, seems to make sense to move to padding?

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
```diff
@@ -14,64 +14,410 @@
...
+ *     // Content to be drawn below status bar (y-axis)
+ * }
+ * ```
+ *
+ * It's also useful when used to draw a scrim which matches the status bar:
+ * ```
+ * Spacer(
+ *     Modifier.statusBarHeight()
+ *         .fillMaxWidth()
+ *         .drawBackground(MaterialTheme.colors.background.copy(alpha = 0.3f)
```
nickbutcher, line 206:
```suggestion
 *         .background(MaterialTheme.colors.background.copy(alpha = 0.3f)
```

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
```diff
@@ -14,64 +14,410 @@
...
+ *     Spacer(Modifier.navigationBarHeight())
+ * }
+ * ```
+ *
+ * It's also useful when used to draw a scrim which matches the navigation bar:
+ * ```
+ * Spacer(
+ *     Modifier.navigationBarHeight()
+ *         .fillMaxWidth()
+ *         .drawBackground(MaterialTheme.colors.background.copy(alpha = 0.3f)
```
nickbutcher, line 236:
```suggestion
 *         .background(MaterialTheme.colors.background.copy(alpha = 0.3f)
```

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
```diff
@@ -14,64 +14,410 @@
...
+
+/**
+ * Declare the preferred width of the content to match the width of the navigation bar,
+ * on the given [side].
+ *
+ * This is very handy when used with `Spacer` to push content inside from any vertical
+ * navigation bars (typically when the device is in landscape):
+ * ```
+ * Row {
+ *     Spacer(Modifier.navigationBarWidth(HorizontalSide.Left))
```
nickbutcher, line 261: How would this handle RTL?

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
```diff
@@ -14,64 +14,410 @@
...
+/**
+ * Allows conditional setting of [insets] on each dimension.
+ */
+private inline fun Modifier.insetsPadding(
+    insets: Insets,
+    left: Boolean = false,
+    top: Boolean = false,
+    right: Boolean = false,
+    bottom: Boolean = false
+) = this then InsetsPaddingModifier(insets, left, top, right, bottom)
```
nickbutcher, line 298: What about making `InsetsPaddingModifier` just accept Px values for l,t,r,b and converting here? That way the modifier wouldn't need to hold the `insets` object and it would only need to do the `if`s once, not on every measure. Actually you could probably move it to using `Modifier.layout`

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
chrisbanes: Updated.

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
chrisbanes: Doesn't need to, insets are always absolute sides

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
chrisbanes: I'm not sure on this one. This is a micro-optimization, and the whole point of this is to move the check from composition to measure.

common-ui-compose/src/main/java/app/tivi/common/compose/insets.kt
chrisbanes: Done.

chrisbanes:
Thanks for the reviews!

PR #640, opened 2020-06-16
Migrate to Hilt

chrisbanes:
This PR moves Tivi away from Dagger-Android to [Hilt](https://dagger.dev/hilt).

Most of the changes are moving modules around or adding `@InstallIn`. Since Tivi uses some Kotlin only modules (i.e. not Android modules), this migration required either `@Module(includes = ...)` them from an Android module, or moving the module.

This PR did require Tivi to migrate away from [MvRx](https://github.com/airbnb/MvRx). Unfortunately MvRx is very opinionated about how its ViewModels are created, which does not work with the new [Hilt AndroidX integration](https://developer.android.com/training/dependency-injection/hilt-jetpack). Luckily Tivi is only using a small subset of what MvRx provides, so I've reimplemented it using Coroutines in `ReduxViewModel`. This was going to be required soon anyway, to be able to use [Jetpack Compose](https://developer.android.com/jetpack/compose)'s upcoming [ViewModel support](https://android-review.googlesource.com/c/platform/frameworks/support/+/1306054).

For me, where Hilt really shines is for testing. No more brittle and hacky Dagger test components just to provide fake dependencies, it's now just a case of `@UninstallModules` for the prod dependencies. See [here](https://dagger.dev/hilt/testing) for more info.

The automatic support for `Worker`s was 👌  too.


buildSrc/src/main/java/app/tivi/buildsrc/dependencies.kt
```diff
@@ -172,20 +170,32 @@ object Libs {
...
+            private const val version = "1.0.0-alpha01"
+            const val work = "androidx.hilt:hilt-work:$version"
+            const val viewmodel = "androidx.hilt:hilt-lifecycle-viewmodel:$version"
+            const val compiler = "androidx.hilt:hilt-compiler:$version"
+        }
     }
 
     object Dagger {
         private const val version = "2.28"
         const val dagger = "com.google.dagger:dagger:$version"
```
manuelvicnt, line 184: You don't need these anymore, Hilt transitively depends on this.

buildSrc/src/main/java/app/tivi/buildsrc/dependencies.kt
```diff
@@ -172,20 +170,32 @@ object Libs {
...
+        object Hilt {
+            private const val version = "1.0.0-alpha01"
+            const val work = "androidx.hilt:hilt-work:$version"
+            const val viewmodel = "androidx.hilt:hilt-lifecycle-viewmodel:$version"
+            const val compiler = "androidx.hilt:hilt-compiler:$version"
+        }
     }
 
     object Dagger {
         private const val version = "2.28"
```
manuelvicnt, line 183: Dagger 2.28 comes with a new artifact for lint checks: `com.google.dagger:dagger-lint`

common-ui-view/src/main/java/app/tivi/TiviFragment.kt
```diff
@@ -16,23 +16,6 @@
...
-    override fun onAttach(context: Context) {
-        AndroidSupportInjection.inject(this)
-        super.onAttach(context)
-    }
-
-    override fun androidInjector(): AndroidInjector<Any> {
-        return androidInjector!!
-    }
-}
+abstract class TiviFragment : Fragment()
```
manuelvicnt, line 21: Maybe remove? It's not doing anything

app/src/main/java/app/tivi/inject/AppModule.kt
```diff
@@ -24,34 +25,38 @@ import androidx.lifecycle.coroutineScope
...
 
-@Module(includes = [AppModuleBinds::class])
+@InstallIn(ApplicationComponent::class)
+@Module(
+    includes = [
+        TraktModule::class,
+        TmdbModule::class
+    ]
+)
 class AppModule {
```
manuelvicnt, line 56: Better performance if modules with `@Provides` are declared as `object` instead of `class`.

Ditto for the rest of the project

data-android/src/test/java/app/tivi/data/DatabaseTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltTestApplication
+import kotlinx.coroutines.test.TestCoroutineScope
+import org.junit.Rule
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@Config(application = HiltTestApplication::class)
+@RunWith(RobolectricTestRunner::class)
```
manuelvicnt, line 29: This can/should be `AndroidJUnit4::class`.

Ditto for any other Robolectric test

buildSrc/src/main/java/app/tivi/buildsrc/dependencies.kt
chrisbanes: I need this for the non-Android modules unfortunately.

app/src/main/java/app/tivi/inject/AppModule.kt
chrisbanes: Thanks for the reminder.

app/src/main/java/app/tivi/home/HomeActivity.kt
```diff
@@ -18,35 +18,30 @@ package app.tivi.home
...
-    @Inject @JvmField var homeNavigationViewModelFactory: HomeActivityViewModel.Factory? = null
+@AndroidEntryPoint
+class HomeActivity : TiviActivity() {
+    private val viewModel: HomeActivityViewModel by viewModels()
 
     private lateinit var binding: ActivityHomeBinding
 
     private var currentNavController: NavController? = null
 
+    @Inject @JvmField var navigator: AppNavigator? = null
```
danysantiago, line 43: You can avoid using the bang operator or null-checking for injected fields knowing you'll access them at the right time if you declare them with `lateinit`:
`@Inject @JvmField lateinit var navigator: AppNavigator`

chrisbanes: Yep, unfortunately there's an issue with `lateinit` with the Kotlin Compiler version used by Compose. This is the only way which currently works 🤦

buildSrc/src/main/java/app/tivi/buildsrc/dependencies.kt
chrisbanes: Thanks, I'll add that in a later PR!

SubhrajyotiSen:
I think the README needs to be updated too
https://github.com/chrisbanes/tivi/blob/main/README.md#android-development

PR #663, opened 2020-07-10
Uses flow's catch instead

jeffreydelooff:
Semantically almost identical but more fluent in terms of Kotlin code.


jeffreydelooff:
@googlebot I fixed it.

chrisbanes:
LGTM, thanks!

PR #669, opened 2020-07-24
Update to SNAPSHOT: #6704997

chrisbanes:


build.gradle
```diff
@@ -29,6 +29,7 @@ buildscript {
         google()
         mavenCentral()
         maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
+        maven { url 'https://dl.bintray.com/kotlin/kotlin-dev' }
```
JoseAlcerreca, line 32: You seem to have a LOT of repositories! Adding comments explaining what they are would be nice.

PR #713, opened 2020-10-16
Remove a load of old code 🗑️ 

chrisbanes:
Lots of view-related code which is no longer being used as the app migrates to Compose 🗑️


axferch89 (commented):
Ok

PR #716, opened 2020-10-19
Add implementation of absolute elevation

chrisbanes:
This is important for show details, because it uses nested surfaces which are each elevated.

Tracking feature request for Compose: https://issuetracker.google.com/171031040

### DIFF
--- a/common-ui-compose/src/main/java/app/tivi/common/compose/Elevation.kt
+++ b/common-ui-compose/src/main/java/app/tivi/common/compose/Elevation.kt
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package app.tivi.common.compose
+
+import androidx.compose.foundation.BorderStroke
+import androidx.compose.material.ElevationOverlay
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Surface
+import androidx.compose.material.contentColorFor
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Providers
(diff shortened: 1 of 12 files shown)


common-ui-compose/src/main/java/app/tivi/common/compose/Elevation.kt
```diff
@@ -0,0 +1,238 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.RectangleShape
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.graphics.compositeOver
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import kotlin.math.ln
+
+val AmbientAbsoluteElevation = staticAmbientOf { 0.dp }
```
nickbutcher, line 55:
```suggestion
val AmbientAbsoluteElevation = ambientOf { 0.dp }
```

As this isn't [highly unlikely](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:compose/runtime/runtime/src/commonMain/kotlin/androidx/compose/runtime/Ambient.kt;l=153) to change

common-ui-compose/src/main/java/app/tivi/common/compose/Elevation.kt
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * A wrapper around [Surface] which works with absolute elevation. This will automatically
```
nickbutcher, line 58:
```suggestion
 * A [Surface] which works with absolute elevation. This will automatically
```

common-ui-compose/src/main/java/app/tivi/common/compose/Elevation.kt
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * An [ElevationOverlay] which supports absolute elevation, taking the current
+ * [AmbientAbsoluteElevation] into account.
+ */
+object AbsoluteElevationOverlay : ElevationOverlay {
+    @Composable
+    override fun apply(color: Color, elevation: Dp): Color {
+        val colors = MaterialTheme.colors
+        val absElevation = AmbientAbsoluteElevation.current + elevation
```
nickbutcher, line 94: Is this "double counting" e.g. if an AbsoluteElevationSurface has an elevation that it passes through to this, hasn't this already been added to the absolute value?

nickbutcher: Ack, yeah a comment would help as I wasn't _sure_ if it was double counting or not lol.

nickbutcher (approved):

PR #737, opened 2020-11-24
Remove AppCompat + MDC

chrisbanes:
Since Tivi has now been converted to Compose, we can now start removing AppCompat & MDC, and other AndroidX view libraries.

## Todo:
- [x] Migrate away from `ConstraintLayout`
- [x] Rewrite multiple back stack to be UI agnostic (not tied to `BottomNavigationView`)
- [x] Remove [OSS Licenses screen](https://developers.google.com/android/guides/opensource) (it depends on AppCompat)
- [x] Migrate Settings back to framework `PreferenceFragment` (AndroidX Preference depends on AppCompat)
- [x] Force exclude AppCompat + MDC dependencies
- [x] Update minify rules to remove views we're not using
- [x] Create native Compose theme
- [x] Fix view theme
- [x] Fix theme setting
- [x] Fix bottom navigation being clipped


axferch89:
Ok

PR #741, opened 2020-12-01
Some small performance improvements and bug fixes

chrisbanes:
- Removed our `asyncOrAwat()` functionality. It's no longer needed now that we're using Store with its multiplexer.
- Fixed the loading counter not removing the loader when the coroutine is cancelled.
- Fixed a crash caused by not filtering for Store's loading state.
- Update Store to 4.0.0

### DIFF
--- a/data/src/main/java/app/tivi/data/StoreExtensions.kt
+++ b/data/src/main/java/app/tivi/data/StoreExtensions.kt
@@ -17,8 +17,11 @@
 package app.tivi.data
 
 import com.dropbox.android.external.store4.Store
+import com.dropbox.android.external.store4.StoreResponse
 import com.dropbox.android.external.store4.fresh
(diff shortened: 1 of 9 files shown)


data/src/main/java/app/tivi/data/StoreExtensions.kt
```diff
@@ -28,3 +31,7 @@ suspend inline fun <Key : Any, Output : Any> Store<Key, Output>.fetch(
     forceFresh -> fresh(key)
     else -> get(key)
 }
+
+fun <T> Flow<StoreResponse<T>>.filterForResult(): Flow<StoreResponse<T>> = filterNot {
```
digitalbuddha, line 35: looks useful, free internet points if you want to add to `StoreResponse`

PR #742, opened 2020-12-01
store stable version tick

digitalbuddha:
Going to yolo this from web interface, will clone and fix if ci fails


chrisbanes:
Haha, thanks! Already done in #741

Issue #747, opened 2021-01-15
main branch is not runnable, error in TraktAuthModule


alishari:
There is an issue while I was running code, In the TraktAuthModule:45 you called AuthorizationServiceConfiguration with three parameters but it has constructors with two or four parameters:
```
e: ...\tivi\trakt\TraktAuthModule.kt: (45, 16): None of the following functions can be called with the arguments supplied: 
public constructor AuthorizationServiceConfiguration(p0: Uri, p1: Uri) defined in net.openid.appauth.AuthorizationServiceConfiguration
public constructor AuthorizationServiceConfiguration(p0: Uri, p1: Uri, p2: Uri?, p3: Uri?) defined in net.openid.appauth.AuthorizationServiceConfiguration
public constructor AuthorizationServiceConfiguration(p0: AuthorizationServiceDiscovery) defined in net.openid.appauth.AuthorizationServiceConfiguration
```

PR #748, opened 2021-01-15
Fix TraktAuthModule compile error

alishari:
As I have described the problem https://github.com/chrisbanes/tivi/issues/747, there was a problem in compiling code. current code called AuthorizationServiceConfiguration with three parameters but it has constructors with two or four parameters.

### DIFF
--- a/trakt-auth/src/main/java/app/tivi/trakt/TraktAuthModule.kt
+++ b/trakt-auth/src/main/java/app/tivi/trakt/TraktAuthModule.kt
@@ -44,8 +44,7 @@ object TraktAuthModule {
     fun provideAuthConfig(): AuthorizationServiceConfiguration {
         return AuthorizationServiceConfiguration(
             Uri.parse("https://trakt.tv/oauth/authorize"),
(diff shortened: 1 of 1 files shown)


alishari:
@googlebot I fixed it.

alishari:
CI task failed due to an error in uploading Crashlytics mapping which is not related to this commint:

```
Execution failed for task ':app:uploadCrashlyticsMappingFileRelease'.
> java.io.IOException: Unknown error while sending file, check network [/home/runner/work/tivi/tivi/app/build/crashlytics/Release/mappings/b7c898f0f8834136ab107d140cd48513.gz; response: 400 HTTP/1.1 400 Bad Request]
```

alishari:
@googlebot I fixed it.

chrisbanes:
Thanks for the PR! Looks like https://github.com/openid/AppAuth-Android/pull/525 caused an API break. I've pinned our AppAuth dependency to a specific revision (ea4dc202407a5ea3d2c9748569374bfd848bc2b2) so we don't hit this in the future.

PR #755, opened 2021-02-02
Update to new transition APIs in Compose 1.0.0-alpha11

chrisbanes:


ui-account/src/main/java/app/tivi/account/AccountUi.kt
```diff
@@ -86,7 +83,7 @@ fun AccountUi(
             ) {
                 @Suppress("DEPRECATION") // TODO: Migrate away from FlowRow
                 FlowRow(
-                    mainAxisAlignment = FlowMainAxisAlignment.End,
+                    mainAxisAlignment = androidx.compose.foundation.layout.FlowMainAxisAlignment.End,
```
nickbutcher, line 86: Import

chrisbanes: Nah, otherwise I need to suppress the whole file which I don't want to do. Using the full class name means I can scope the suppression to just here.

ui-showdetails/src/main/java/app/tivi/showdetails/details/ShowDetails.kt
```diff
@@ -367,7 +359,7 @@ private fun ShowDetailsScrollingContent(
             }
 
             seasons.forEach {
-                SeasonWithEpisodesRow(
+                seasonWithEpisodesRow(
```
nickbutcher, line 362: I get that this *isn't* a composable function but the naming feels wrong. Might be worth a wider discussion on this?

ui-showdetails/src/main/java/app/tivi/showdetails/details/ShowDetails.kt
```diff
@@ -441,23 +433,46 @@ private fun OverlaidStatusBarAppBar(
     LogCompositions("OverlaidStatusBarAppBar")
 
     Column(modifier) {
-        val props = transition(
-            definition = statusBarTransitionDef,
-            initState = false,
-            toState = showAppBar
+        val transition = updateTransition(showAppBar)
+
+        val elevation by transition.animateDp { value ->
```
nickbutcher, line 438: Name the param?

ui-showdetails/src/main/java/app/tivi/showdetails/details/ShowDetails.kt
```diff
@@ -441,23 +433,46 @@ private fun OverlaidStatusBarAppBar(
     LogCompositions("OverlaidStatusBarAppBar")
 
     Column(modifier) {
-        val props = transition(
-            definition = statusBarTransitionDef,
-            initState = false,
-            toState = showAppBar
+        val transition = updateTransition(showAppBar)
```
nickbutcher, line 436: What about extracting this transition to a function, returning some `data class BarTransition(val offset, val alpha)`?

chrisbanes: Currently each prop state means that recomposition is next to a minimum, if we extract to group class: 1) We will be wasting a lot of objects. 2) trigger recomposition a lot more.

I think the current code is fine. WDYT?

ui-showdetails/src/main/java/app/tivi/showdetails/details/ShowDetails.kt
chrisbanes: We then get into shadowed names. Not sure what is better here tbh.

ui-showdetails/src/main/java/app/tivi/showdetails/details/ShowDetails.kt
chrisbanes: In fact, 2) isn't an issue since they'll be `equals()`

ui-showdetails/src/main/java/app/tivi/showdetails/details/ShowDetails.kt
```diff
@@ -441,60 +436,81 @@ private fun OverlaidStatusBarAppBar(
...
                 content = content,
             )
         }
     }
 }
 
-private val OffsetYKey = DpPropKey()
-private val AlphaKey = FloatPropKey()
+@Composable
+private fun updateOverlaidStatusBarAppBarTransition(
```
nickbutcher, line 464: I like where this ended up, moving the transition code out of the UI code.

ui-showdetails/src/main/java/app/tivi/showdetails/details/ShowDetails.kt
```diff
@@ -360,24 +354,25 @@ private fun ShowDetailsScrollingContent(
...
         if (seasons.isNotEmpty()) {
-            spacerItem(8.dp)
+            itemSpacer(8.dp)
 
             item {
                 Header(stringResource(R.string.show_details_seasons))
             }
 
-            seasons.forEach {
+            items(seasons) {
```
nickbutcher, line 363: nit: name param

ui-showdetails/src/main/java/app/tivi/showdetails/details/ShowDetails.kt
nickbutcher: I like where this ended up

nickbutcher (approved):
One nit but otherwise LGTM. Thanks for the changes.

PR #756, opened 2021-02-04
Move away from storing dependencies in buildSrc

chrisbanes:
It causes a cache invalidation whenever the dependency version changes, and it makes it difficult for tools to process the files.

I tried to keep the structure similar to before, to reduce changes in the module `build.gradle` files.


versions.gradle
```diff
@@ -0,0 +1,175 @@
...
+
+versions.coil = "1.1.1"
+libs.Coil = [
+    coil: "io.coil-kt:coil:$versions.coil"
+]
+
+versions.roomigrant = "master-SNAPSHOT"
+libs.Roomigrant = [
+    library: "com.github.chrisbanes.Roomigrant:RoomigrantLib:$versions.roomigrant",
+    compiler: "com.github.chrisbanes.Roomigrant:RoomigrantCompiler:$versions.roomigrant",
```
itsandreramon, line 191: Where can I find this version of Roomigrant?

chrisbanes: https://github.com/chrisbanes/roomigrant

It's just ports it over to AndroidX

PR #761, opened 2021-02-16
Migration to Navigation Compose

chrisbanes:
This PR removes (nearly) all Fragments from Tivi*, and migrates everything to Compose Navigation. There are no animations (yet), but that will come.

\* _The only fragment left is a `PreferenceFragment` which I likely won't migrate over to Compose for a while._


ui-episodedetails/src/main/java/app/tivi/episodedetails/EpisodeDetailsViewModel.kt
```diff
@@ -54,7 +55,12 @@ class EpisodeDetailsViewModel @AssistedInject constructor(
     private val removeEpisodeWatch: RemoveEpisodeWatch,
     private val logger: Logger,
     private val snackbarManager: SnackbarManager
-) : ReduxViewModel<EpisodeDetailsViewState>(initialState) {
+) : ReduxViewModel<EpisodeDetailsViewState>(
+    EpisodeDetailsViewState(
+        // The string "episodeId" is the name of the argument in the route
+        episodeId = savedStateHandle.get<String>("episodeId")!!.toLong()
```
ianhanniballake, line 61: Note that [defining a type on your argument](https://developer.android.com/jetpack/compose/navigation#nav-with-args) will mean this is already a Long, avoiding the need to cast.

chrisbanes: Hmmm I got a `ClassCastException` when I tried that earlier, which is why I added the explicit conversion. I'll double check tomorrow.

chrisbanes: Ohhh, just read the link. Makes sense, I'll move over to that.

chrisbanes: Done in #778

PR #788, opened 2021-04-23
Tidy up ViewState collection from ViewModels

chrisbanes:
Using `flowWithLifecycle()` to reduce resource usage when the app is put into the background

### DIFF
--- a/common-ui-compose/src/main/java/app/tivi/common/compose/Flow.kt
+++ b/common-ui-compose/src/main/java/app/tivi/common/compose/Flow.kt
@@ -18,12 +18,12 @@ package app.tivi.common.compose
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.State
-import androidx.compose.runtime.collectAsState
-import androidx.compose.runtime.remember
+import androidx.compose.runtime.produceState
 import androidx.compose.ui.platform.LocalLifecycleOwner
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.flowWithLifecycle
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
 
 @Composable
 fun <T> Flow<T>.collectAsStateWithLifecycle(
@@ -31,11 +31,12 @@ fun <T> Flow<T>.collectAsStateWithLifecycle(
     minActiveState: Lifecycle.State = Lifecycle.State.STARTED,
 ): State<T> {
     val lifecycleOwner = LocalLifecycleOwner.current
-
-    return remember(this, lifecycleOwner) {
-        this@collectAsStateWithLifecycle.flowWithLifecycle(
+    return produceState(initialValue = initial, this, lifecycleOwner) {
+        flowWithLifecycle(
             lifecycle = lifecycleOwner.lifecycle,
             minActiveState = minActiveState
-        )
-    }.collectAsState(initial = initial)
+        ).collect {
+            value = it
+        }
+    }
 }
--- a/common-ui-compose/src/main/java/app/tivi/common/compose/Flow.kt
+++ b/common-ui-compose/src/main/java/app/tivi/common/compose/Flow.kt
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("NOTHING_TO_INLINE")
+
 package app.tivi.common.compose
 
 import androidx.compose.runtime.Composable
@@ -40,3 +42,11 @@ fun <T> Flow<T>.collectAsStateWithLifecycle(
         }
     }
 }
+
+@Composable
+inline fun <T> Flow<T>.flowWithLocalLifecycle(
+    minActiveState: Lifecycle.State = Lifecycle.State.STARTED
+): Flow<T> = flowWithLifecycle(
+    lifecycle = LocalLifecycleOwner.current.lifecycle,
+    minActiveState = minActiveState
+)
--- a/common-ui-compose/src/main/java/app/tivi/common/compose/Flow.kt
+++ b/common-ui-compose/src/main/java/app/tivi/common/compose/Flow.kt
@@ -19,34 +19,20 @@
 package app.tivi.common.compose
 
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.State
-import androidx.compose.runtime.produceState
+import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.LocalLifecycleOwner
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.flowWithLifecycle
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.collect
 
 @Composable
-fun <T> Flow<T>.collectAsStateWithLifecycle(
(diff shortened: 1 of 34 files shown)


common-ui-view/src/main/java/app/tivi/ui/SnackbarManager.kt
```diff
@@ -33,6 +36,23 @@ class SnackbarManager @Inject constructor() {
     private val pendingErrors = MutableSharedFlow<UiError>(extraBufferCapacity = 3)
     private val removeErrorSignal = MutableSharedFlow<Unit>(extraBufferCapacity = 1)
 
+    val flow: Flow<UiError?> = flow {
```
manuelvicnt, line 39: Can you add comments about what's the purpose of this?

common-ui-view/src/main/java/app/tivi/ui/SnackbarManager.kt
```diff
@@ -33,6 +36,23 @@ class SnackbarManager @Inject constructor() {
     private val pendingErrors = MutableSharedFlow<UiError>(extraBufferCapacity = 3)
```
manuelvicnt, line 36: Does this work as you expect? Even if there are no collectors, messages will be sent and they could be potentially lost. This `extraBufferCapacity` is only helpful when `SharedFlow` has slow subscribers. `emit()` will suspend until all collectors receive the event.

ui-episodedetails/src/main/java/app/tivi/episodedetails/EpisodeDetailsViewModel.kt
```diff
@@ -55,29 +56,38 @@ internal class EpisodeDetailsViewModel @Inject constructor(
...
-        viewModelScope.launch {
-            observeEpisodeWatches.observe()
-                .onStart { emit(emptyList()) }
-                .collect { updateFromEpisodeWatches(it) }
-        }
+    val state = combine(
+        observeEpisodeDetails.observe(),
+        observeEpisodeWatches.observe(),
+        loadingState.observable,
+        snackbarManager.flow.onStart { emit(null) },
```
manuelvicnt, line 71: Why do you need to emit `null` on start? `snackbarManager.flow` already does that for you, right?

ui-episodedetails/src/main/java/app/tivi/episodedetails/EpisodeDetailsViewModel.kt
```diff
@@ -55,29 +56,38 @@ internal class EpisodeDetailsViewModel @Inject constructor(
...
+            watches = episodeWatches,
+            canAddEpisodeWatch = episodeDetails.episode?.firstAired?.let {
+                it.isBefore(OffsetDateTime.now())
+            } ?: true,
+            refreshing = refreshing,
+            error = error,
+        )
+    }.stateIn(
+        scope = viewModelScope,
+        started = WhileSubscribed(5000),
```
manuelvicnt, line 86: This is only helpful if you're collecting the flow using the new `addRepeatingJob` API or you're manually cancelling the job/scope when the view goes to the background. This won't work if you're using `launchWhenStarted`.

Which one are you using?

chrisbanes: Hmmm, these are collected from Compose via `collectAsState()`. Oh well.

manuelvicnt: Ah! Then, yes, you're having the same problems unless you remember a `Flow.flowWithLifecycle` as you can read here: https://medium.com/androiddevelopers/a-safer-way-to-collect-flows-from-android-uis-23080b1f8bda

Relevant code snippet: https://gist.github.com/manuelvicnt/6629b53e33b97c127508227ed6e0bee4#file-flowwithlifecyclecompose-kt

common-ui-view/src/main/java/app/tivi/ui/SnackbarManager.kt
manuelvicnt: The following code buffers UiErrors in the channel when there are no collectors

```
private val _pendingErrors = Channel<UiError>(3, DROP_OLDEST)
val pendingErrors = _navigationEvents.receiveAsFlow().shareIn(externalScope, WhileSubscribed())
```

ui-episodedetails/src/main/java/app/tivi/episodedetails/EpisodeDetailsViewModel.kt
chrisbanes: Migrated to `flowWithLifecycle()` 👍

ui-episodedetails/src/main/java/app/tivi/episodedetails/EpisodeDetailsViewModel.kt
chrisbanes: Good call, removed.

common-ui-view/src/main/java/app/tivi/ui/SnackbarManager.kt
chrisbanes: Thanks, went back to using a Channel. Wasn't sure that we need the `shareIn()`. PTAL

common-ui-view/src/main/java/app/tivi/ui/SnackbarManager.kt
chrisbanes: Added.

manuelvicnt (approved):

common-ui-view/src/main/java/app/tivi/ui/SnackbarManager.kt
manuelvicnt: Looks good 👍  thank you!

data-android/src/main/java/app/tivi/data/TiviRoomDatabase.kt
```diff
@@ -62,7 +62,6 @@ import app.tivi.data.views.FollowedShowsWatchStats
     ],
     version = 26,
     autoMigrations = [
-        AutoMigration(from = 23, to = 24),
```
chrisbanes, line 65: No idea why the build doesn't like it. It's not a big issue to remove this anyway.

common-ui-compose/src/main/java/app/tivi/common/compose/Flow.kt
```diff
@@ -0,0 +1,52 @@
...
+            lifecycle = lifecycleOwner.lifecycle,
+            minActiveState = minActiveState
+        ).collect {
+            value = it
+        }
+    }
+}
+
+@Composable
+inline fun <T> Flow<T>.flowWithLocalLifecycle(
```
manuelvicnt, line 47: Does this need to be remembered for paging? If so, you might want to call it `inline fun rememberFlowWithLocalLifecycle(flow)

chrisbanes: It actually works fine right now, but I like the look of `rememberFlowWithLifecycle` so migrated all calls to it

PR #803, opened 2021-05-14
Setup refreshVersions plugin

### LINKED ISSUE #802: Use gradle refreshVersions?
Short story: if you are interested, I can submit a pull request to setup gradle refreshVersions in the project in order to make it easy to find available dependencies updates
https://jmfayard.github.io/refreshVersions/update-dependencies/

Longer story:

[gradle refreshVersions](https://jmfayard.github.io/refreshVersions/) is a Gradle plugin and my attempt at making managing dependencies faster and less error-prone. We just released a stable version 0.10.0

I'm trying on multiple oppen-source projects to make sure we are covering various use cases well.

What I found interesting in TiVi is that it's using versions catalogs, which is an interesting but currently rarely used feature. It overlaps with gradle refreshVersions - the adding dependency part - but can be combined with it as well and I'm interested to find out how well that works.

jmfayard:
Closes #802

Documentation available at
https://jmfayard.github.io/refreshVersions/

Co-Authored-By: Louis CAD <louis.cognault@gmail.com>


jmfayard:
@googlebot I consent

jmfayard:
@googlebot I consent.

LouisCAD:
@googlebot I consent.

lordcodes:
@jmfayard Personally, I think this is a bit of a step-back from what was already there. It was using the new Gradle built-in approach for managing dependencies and their versions.

After these changes all the versions are managed externally in a properties file. I think you may want to consider altering the refreshVersions plugin so that it can read and update the Gradle dependencies TOML file. Once Gradle 7.0 and the dependencies TOML become more stable I think projects will be looking to update to it.

jmfayard:
@lordcodes 
> I think you may want to consider altering the refreshVersions plugin so that it can read and update the Gradle dependencies TOML file
It's on the roadmap, but not urgent because the version catalogs are neither stable nor widely used.

I don't think it's a step back: Gradle does not have a built-in way to lookup for available updates and I think refreshVersions is very valuable there.

lordcodes:
@jmfayard

> I don't think it's a step back: Gradle does not have a built-in way to lookup for available updates and I think refreshVersions is very valuable there.

100% agree about it helping with looking up available updates and applying them.

I was referring to how it moved the versions out of the file that is the new Gradle built-in approach and using it's own file to store the versions.

jmfayard:
> 100% agree about it helping with looking up available updates and applying them.

Thanks for clarifying :)

> I was referring to how it moved the versions out of the file that is the new Gradle built-in approach and using it's own file to store the versions.

Like I said, it's on the roamp. refreshVersions had the idea to move the versions away from the code and into a proper file format before the versions catalog were a thing

lordcodes:
@jmfayard Okay sounds great! I think the combination of Gradle dependencies TOML file and refreshVersions to find new versions and apply them would be very powerful, so glad to hear it's on the roadmap.

chrisbanes:
I'm with @lordcodes. I'm not sure that I like the two file approach as it makes it more difficult to maintain (even if it auto updates).

Version catalogs provide a [settings API](https://docs.gradle.org/current/userguide/platforms.html) allowing plugins + builds to change the dependencies as required, I think that is a valid path forward for the `refreshVersions` plugin.

jmfayard:
I agree, but we will still have `versions.properties` in any case for people stuck on Gradle 6

> Version catalogs provide a settings API allowing plugins + builds to change the dependencies as required,

That's already supported, for example with Firebase https://github.com/jmfayard/refreshVersions/blob/b4f1dd082039781f3191ee425a82c84e432d076f/plugins/dependencies/src/main/kotlin/dependencies/Firebase.kt#L19-L29

LouisCAD:
Yup, I spent hours (days?) hand crafting all the AndroidX, Firebase and a bunch of other dependency notations that are embedded into refreshVersions.

I still prefer the API it provides over what Gradle's versions catalog allows, but we're still considering adding first-class support for that Gradle 7+ feature.

jmfayard:
@chrisbanes 
`refreshVersions` now supports adding available updates (as comment) directly inside `gradle/libs.versions.toml` so that there is no need to do any big modification in your build anymore.

```properties
## gradle/libs.versions.toml
## To find automatically available updates, run
## ./gradlew refreshVersions
[versions]

junit-junit = "4.12"
##       ⬆ = "4.13"  
##       ⬆ = "4.13.1"

[libraries]

junit = { group = "junit", name = "junit", version.ref = "junit-junit" }

guava = "com.google.guava:guava:30.0-jre"
##                          ⬆ :31.0-android"  
##                          ⬆ :31.0-jre"
```

If interested, have a look here

https://github.com/jmfayard/refreshVersions/wiki/RefreshVersions-%E2%99%A5%EF%B8%8F-Gradle-Version-Catalog#if-you-already-use-a-versions-catalog

Feedback and bugs should go here @lordcodes  @LouisCAD 

https://github.com/jmfayard/refreshVersions/discussions/592
cc @lordcodes  @LouisCAD

PR #806, opened 2021-05-19
[WIP] Play around with creating our own 'ViewModel'

chrisbanes:
See this thread: https://twitter.com/joreilly/status/1395103281203093511

This is just a proof of concept, it is definitely  not perfect. It does not have all of the characteristics of AAC ViewModels, but keeps the concept of a ‘state holder’.

**Don't copy this in your apps.** It's purely an experiment.


common-ui-compose/src/main/java/app/tivi/common/compose/StateModel.kt
```diff
@@ -0,0 +1,47 @@
...
+ * A ViewModel-like class which can be used from composables.
+ *
+ * It implements [RememberObserver] which allows it to hold a [CoroutineScope], and cancel the
+ * scope's job once the model has been 'cleared'.
+ */
+abstract class StateModel : RememberObserver {
+    /**
+     * A [CoroutineScope] which can be used similar to `viewModelScope`
+     */
+    protected val coroutineScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
```
manuelvicnt, line 35: Nit: `Dispatchers.Main.immediate + SupervisorJob()` is what `viewModelScope` actually uses

ui-episodedetails/src/main/java/app/tivi/episodedetails/EpisodeDetails.kt
```diff
@@ -101,20 +102,47 @@ import com.google.accompanist.coil.rememberCoilPainter
...
 fun EpisodeDetails(
+    id: Long,
     navigateUp: () -> Unit,
 ) {
+    // TODO: handle ContextWrappers and unwrap as necessary
+    val context = LocalContext.current as Activity
+
     EpisodeDetails(
-        viewModel = hiltViewModel(),
+        viewModel = remember(id) {
```
manuelvicnt, line 122: Good usage of Hilt here 👍  I was wondering how we could propagate `onForgotten` calls to the `StateModel`. 

How safe would it be if apart from the `remember(id)` to create a `DisposableEffect(id)` that does that?

```
val viewModel = remember(id) { /* ... */ }
DisposableEffect(id) {
    onDispose { viewModel.onForgotten() }
}
```

We'll need to investigate implications when multi-threaded compositions take place. But for now... I think that should be safe to use?

chrisbanes: Yep. This is exactly how `rememberCoroutineScope()` is implemented. I also use this pattern in Accompanist.

manuelvicnt: Ah, then `DisposableEffect` is not needed. TIL `RememberObserver` which already does all of that for you. That's great!

chrisbanes: It’s super powerful, I’ve been meaning to ping you and about doc-ing this pattern more.

common-ui-compose/src/main/java/app/tivi/common/compose/StateModel.kt
chrisbanes: Fixed.

manuelvicnt (approved):
Like it!!!

chrisbanes:
@manuelvicnt WDYT to `ViewModelStore`? It's what we were chatting about yesterday and seems to work nicely.

app/src/main/java/app/tivi/ViewModelStore.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        // If not we'll create a new ViewModel
+
+        val viewModelScope = createChildCoroutineScope()
+
+        val flow = flow<T> {
+            // This is a 'fake' flow which doesn't emit anything. The actual ViewModel is
+            // created in the stateIn() below. We suspend the flow until its cancelled, which
+            // happens when either the stateIn `WhileSubscribed`, or the parent scope is cancelled
+            awaitCancellation()
```
manuelvicnt, line 72: You should be able to use an `emptyFlow()` here. It's fine if this flow completes, actually, that's better as it won't remain in memory. What we care about is the `StateFlow` that gets created with `stateIn` and its cached value. That `StateFlow` will be in memory until `coroutineScope` gets cancelled or there are no GC references to it.

app/src/main/java/app/tivi/ViewModelStore.kt
```diff
@@ -0,0 +1,106 @@
...
+import kotlinx.coroutines.isActive
+
+/**
+ * TODO
+ */
+@Stable
+internal class ViewModelStore(
+    coroutineScope: CoroutineScope,
+) {
+    private val viewModelMap = HashMap<Any, ViewModelStoreEntry>()
```
manuelvicnt, line 40: We'll need to protect this map against concurrent reads/writes

app/src/main/java/app/tivi/ViewModelStore.kt
manuelvicnt: I assume you've done this because of `viewModelScope.cancel()`. What about having that logic on the `StateFlow`? You can move the operator below `stateIn` and do the same

app/src/main/java/app/tivi/ViewModelStore.kt
```diff
@@ -0,0 +1,106 @@
...
+            SupervisorJob(coroutineScope.coroutineContext[Job])
+    )
+
+    /**
+     * Retrieve or create a ViewModel.
+     */
+    fun <T : Any> viewModelFlow(
+        key: Any,
+        create: (CoroutineScope) -> T
+    ): StateFlow<T> {
```
manuelvicnt, line 56: What about changing this API to be `getViewModel(key, create): ViewModel<T>`? What's nice about `StateFlow` is that you can access its `value` synchronously. We don't want to get ViewModel updates from the consumer side (there won't be updates), returning `flow.value` from this function should be enough. Even better, the API is easier to consume and more readable

app/src/main/java/app/tivi/ViewModelStore.kt
chrisbanes: Yep, this is due to the `onComplete {}` block. Setting the `onComplete {}` on the `StateFlow` doesn't work, as the StateFlow always has a collector then, which breaks the `WhileSubscribed`.

app/src/main/java/app/tivi/ViewModelStore.kt
chrisbanes: We need the collection of the StateFlow to be in the UI, so that the `WhileSubscribed` works as expected. We could provide a `@Composable` function which returns a `ViewModel` though.

app/src/main/java/app/tivi/ViewModelStore.kt
```diff
@@ -0,0 +1,106 @@
...
+            // created in the stateIn() below. We suspend the flow until its cancelled, which
+            // happens when either the stateIn `WhileSubscribed`, or the parent scope is cancelled
+            awaitCancellation()
+        }.onCompletion {
+            // The flow has been cancelled, so we need to also cancel the
+            // ViewModel's coroutine scope
+            viewModelScope.cancel()
+        }.stateIn(
+            scope = coroutineScope,
+            started = SharingStarted.WhileSubscribed(5000),
```
manuelvicnt, line 79: This should be `SharingStartedWhileSubscribed(5000, replayExpirationMillis = 0)`.

Once `viewModelScope.cancel()` is called, the ViewModel is unusable, we'd need to create a new instance of it. However, `replayExpirationMillis` defaults to `Long.MAX_VALUE` which means that the same instance of the ViewModel will be _always_ used. We don't want that as its `coroutineScope` will be already cancelled. 

Ideally, we'd create a whole new ViewModel with a new `viewModelScope`.

app/src/main/java/app/tivi/ViewModelStore.kt
```diff
@@ -0,0 +1,106 @@
...
+        // First check if we have a 'cached' ViewModel which is still active
+        val cached = viewModelMap[key]?.takeIf { it.isActive }
+        if (cached != null) {
+            @Suppress("UNCHECKED_CAST")
+            return cached.flow as StateFlow<T>
+        }
+
+        // If not we'll create a new ViewModel
+
+        val viewModelScope = createChildCoroutineScope()
```
manuelvicnt, line 66: We need to make this a `var` to reassign it whenever `onCompletion` is called. See comment below

app/src/main/java/app/tivi/ViewModelStore.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        val flow = flow<T> {
+            // This is a 'fake' flow which doesn't emit anything. The actual ViewModel is
+            // created in the stateIn() below. We suspend the flow until its cancelled, which
+            // happens when either the stateIn `WhileSubscribed`, or the parent scope is cancelled
+            awaitCancellation()
+        }.onCompletion {
+            // The flow has been cancelled, so we need to also cancel the
+            // ViewModel's coroutine scope
+            viewModelScope.cancel()
```
manuelvicnt, line 76:
```
viewModelScope.cancel()
viewModelScope = createChildCoroutineScope()
```

If we need to create another ViewModel after the view's been unsubscribed, we cannot reuse `viewModelScope` as it's been cancelled

app/src/main/java/app/tivi/ViewModelStore.kt
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * TODO
+ */
+@Stable
+internal class ViewModelStore(
+    coroutineScope: CoroutineScope,
+) {
+    private val viewModelMap = HashMap<Any, ViewModelStoreEntry>()
+
```
manuelvicnt, line 41: Do we want to create a maintenance job to clean entries from the `HashMap`?

```
init {
    coroutineScope.launch(ioDispatcher) {
        // go through all ViewModels every so often and remove inactive ones 
    }
}
```

app/src/main/java/app/tivi/ViewModelStore.kt
manuelvicnt: Actually, once you cancel the coroutineScope.. you might also want to remove it from the `HashMap`?

app/src/main/java/app/tivi/ViewModelStore.kt
manuelvicnt: Actually, this might not be necessary if we remove the entry in the `onCompletion` block below

app/src/main/java/app/tivi/ViewModelStore.kt
manuelvicnt: Right! That's true. Yeah, probably we need a `@Composable` function to hide this complexity away

app/src/main/java/app/tivi/ViewModelStore.kt
manuelvicnt: Yes sorry! I've been adding comments as I was going through the code. Everything makes much more sense now 👍  (I've been removing some of my comments that don't make sense, sorry)

app/src/main/java/app/tivi/ViewModelStore.kt
chrisbanes: Done

app/src/main/java/app/tivi/ViewModelStore.kt
chrisbanes: Removed the entry instead

app/src/main/java/app/tivi/ViewModelStore.kt
chrisbanes: Done

manuelvicnt (approved):
Love how this ended up looking!!!

ChanSek:
@chrisbanes, any reason for this branch not being merged to `main`? Is there some cons for not using AAC ViewModel which I am missing?

chrisbanes:
It was just an experiment. There are no cons to not using AAC VMs, I just didn't want to maintain this myself.

PR #839, opened 2021-08-01
Remove obsolete sdk check

wiryadev:


chrisbanes:
I'm actually going to go back to a minkSdkVersion of 23 once https://issuetracker.google.com/issues/194289155 is fixed, so I keep those checks for now.

Thanks anyway

PR #852, opened 2021-08-21
Fix flipped fav icon (UX)

theapache64:
I've noticed the favorite icon behaves differently (showing enabled state for disabled and vice versa)


chrisbanes (approved):

chrisbanes:
Thanks!

PR #880, opened 2021-12-23
Add a Spacer in EpisodeDetails.kt

f2janyway:
When episode watches fill over the (bottom)`navigationBar`, even it has a`Spacer(Modifer.navigationBarsHeight())` , last watch is hide  behind `navigationBar`. 
After add a `Spacer(Modifer.navigationBarsHeight())`, watches fit right.
But I don't know why a `Spacer(Modifer.navigationBarsHeight())` have not effect well.
I think there is a more clear way than this.


chrisbanes:
Thanks for the contribution. I've fixed it in #882

PR #891, opened 2021-12-31
UIs now only receive a single UiMessage to display

chrisbanes:
Follow on from #889. 

Previously each view state would contain a list of the queued up messages. This means that the UI is responsible for deciding what message to display, which is a bit too much logic in the UI for my liking.

This PR changes this so each view state contains a single `UiMessage?`. The queuing is moved to `UiMessageManager` which means that all of the logic is still in the VM. We could later add some kind of message prioritisation, skip messages, etc later without any changes to the UI.

WDYT @manuelvicnt?

### DIFF
--- a/common-ui-view/src/main/java/app/tivi/api/UiMessage.kt
+++ b/common-ui-view/src/main/java/app/tivi/api/UiMessage.kt
@@ -16,9 +16,10 @@
 
 package app.tivi.api
 
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.StateFlow
-import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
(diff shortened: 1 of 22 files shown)


manuelvicnt:
Hi Chris! Moving that logic out of the UI LGTM, in fact, that's something we do in some Jet* apps as well :D we move it to the plain state holder instead of the manager... but it's pretty much the same thing. Thanks!

SteinerOk:
@chrisbanes, what the reason to use `Mutex` in `UiMessageManager` instead of `update` method in `MutableStateFlow` (like [here](https://github.com/Ahmed-Mahmoud7/AndroidAutomationEspresso/blob/main/app/src/main/java/com/example/jetsnack/model/SnackbarManager.kt)) ?

chrisbanes:
The reason is because of potential multi-threading. Each call to `emitMessage()` is additive, in that it takes the current value, adds to it, and then emits it back to the `StateFlow`. We need to make sure that nothing changes `_messages.value` at the same time.

PR #900, opened 2022-02-18
rememberStateWithLifecycle utility

yschimke:
Minor cleanup and simplification of initial value which exists from StateFlow.

I was using your implementation, but wanted to avoid passing in an extra initial value,
and figured worth pushing back to here.  No problem if not desired, but I'd love to here your
take if so.


chrisbanes (approved):
LGTM, thanks!

PR #919, opened 2022-06-01
Fix AccountUi lag

janbina:
There is a massive UI lag when opening the AccountUi dialog, which is caused by the AppAuth's `AuthorizationService`:

> AuthorizationService will create a bound service connection to the browser in order to perform a "warm up" that makes the custom tab load faster. As such, you should create your AuthorizationService instance early in your activity's lifecycle to benefit from this warmup, and not at the moment at which you click "sign in".\
 [https://github.com/openid/AppAuth-Android/issues/328](https://github.com/openid/AppAuth-Android/issues/328)

As of now, `ActivityTraktAuthManager::buildLoginIntent` is called right after AccountUi dialog is opened.\
I modified `LoginTrakt` and `TraktAuthManager` such that `buildLoginIntent` is only called after user clicks the login button and the `ActivityTraktAuthManager` such that `AuthorizationService` is not initialized lazily.


chrisbanes (approved):
Great detective work! Thanks

PR #927, opened 2022-07-11
Fix baseline profile setup

keyboardsurfer:
Main issue was the missing kotlin gradle plugin.

Also added a baseline profile generator.


chrisbanes (approved):

chrisbanes:
Thanks for tracking this down @keyboardsurfer!

PR #938, opened 2022-08-12
Optimize CI a bit

Goooler:
Use [gradle-build-action](https://github.com/gradle/gradle-build-action) to speed up build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 11
+          java-version: 17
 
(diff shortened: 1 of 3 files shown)


Goooler:
Sorry, I just see you enabled gradle-enterprise but without `publishAlways`, if we should add this?

https://github.com/chrisbanes/tivi/blob/9972e02686ebaffd54052ca874c4d917f0cf094b/settings.gradle#L22-L27

chrisbanes:
> Sorry, I just see you enabled gradle-enterprise but without `publishAlways`, if we should add this?
> 
> 
> 
> https://github.com/chrisbanes/tivi/blob/9972e02686ebaffd54052ca874c4d917f0cf094b/settings.gradle#L22-L27

Ah, yes please

PR #941, opened 2022-08-23
Try out KotlinX Immutable Collections

chrisbanes:
WIP


base/src/main/java/app/tivi/extensions/ImmutableCollections.kt
arkivanov, line 1: Wondering, why there is still "Copyright 2022 Google LLC"? :-)

PR #947, opened 2022-09-14
Spotless 6.11.0 & Ktlint 0.47.1

Goooler:
If we should enable `trailing_comma` like https://github.com/cashapp/molecule/commit/7c221efdcf2a7de4f94e79673f6e99827b592de1


data/src/main/java/app/tivi/data/entities/Episode.kt
```diff
@@ -41,7 +41,7 @@ import org.threeten.bp.OffsetDateTime
     ]
 )
 data class Episode(
-    @PrimaryKey(autoGenerate = true) @ColumnInfo(name = "id")
```
chrisbanes, line 44: What's the thinking behind this change? R8 obfuscation *might* rename the field, which would then break the mapping.

Goooler: Room will keep fields in Entity classes, and uses the field names as column names by default.

Goooler: We can even remove all redundant `@ColumnInfo`s.

chrisbanes: I prefer the explicit `@ColumnInfo` as it makes everything very clear. Please revert this commit.

Goooler: Done.

chrisbanes: Thanks!

PR #950, opened 2022-09-19
Add `unsafeLazy` to simplify thread unsafe lazy initializations

Goooler:
Like: https://github.com/runningcode/kotlin-dsl/blob/master/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/unsafeLazy.kt

### DIFF
--- a/base/src/main/java/app/tivi/extensions/Lazy.kt
+++ b/base/src/main/java/app/tivi/extensions/Lazy.kt
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


base/src/main/java/app/tivi/extensions/Lazy.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package app.tivi.extensions
+
+fun <T> unsafeLazy(initializer: () -> T): Lazy<T> = lazy(LazyThreadSafetyMode.NONE, initializer)
```
chrisbanes, line 19: We can make this an `inline` function too

chrisbanes (commented):
Looks great overall

base/src/main/java/app/tivi/extensions/Lazy.kt
Goooler: But also need to add `noinline` on `initializer`, should we do?

chrisbanes: Yeah sounds good. This function compiles down to what we had previously, just with a nicer API

chrisbanes (approved):
Thanks!

PR #951, opened 2022-09-21
[WIP] Try out sergei-lapin/napt

chrisbanes:
Migrates to @sergei-lapin's [napt](https://github.com/sergei-lapin/napt), to replace kapt.

- Had to disable Hilt's aggregating task as it doesn't seem to work with napt
- gitignore NaptTrigger

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,7 @@ org.eclipse.buildship.core.prefs
 .classpath
 .project
 bin/
+
+# napt triggers
+# https://github.com/sergei-lapin/napt
+NaptTrigger.java
--- a/api/tmdb/build.gradle.kts
+++ b/api/tmdb/build.gradle.kts
@@ -17,7 +17,7 @@
 
 plugins {
     id("kotlin")
-    alias(libs.plugins.kotlin.kapt)
+    alias(libs.plugins.kotlin.napt)
     alias(libs.plugins.android.lint)
 }
 
@@ -31,5 +31,5 @@ dependencies {
 
     implementation(libs.okhttp.okhttp)
 
-    kapt(libs.dagger.compiler)
+    annotationProcessor(libs.dagger.compiler)
 }
--- a/api/trakt-auth/build.gradle.kts
+++ b/api/trakt-auth/build.gradle.kts
@@ -18,7 +18,7 @@
 plugins {
     alias(libs.plugins.android.library)
     alias(libs.plugins.kotlin.android)
-    alias(libs.plugins.kotlin.kapt)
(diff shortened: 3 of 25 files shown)


sergei-lapin:
Hi, @chrisbanes ! Right now, I think, you can get this working via disabling Hilt's aggregating task:
```
hilt {
    enableAggregatingTask = false
}
```
while I'm thinking of a better way to avoid this kind of errors

chrisbanes:
> Hi, @chrisbanes ! Right now, I think, you can get this working via disabling Hilt's aggregating task:
> 
> ```
> hilt {
>     enableAggregatingTask = false
> }
> ```
> 
> while I'm thinking of a better way to avoid this kind of errors

Hi @sergei-lapin, I briefly tried that but it still didn't work. I'll try again though

build.gradle
```diff
@@ -93,9 +93,10 @@ subprojects {
         }
     }
 
-    plugins.withId(rootProject.libs.plugins.hilt.get().pluginId) {
-        hilt.enableAggregatingTask = true
-    }
+    // Disable this whilst trying out napt #951
+    // plugins.withId(rootProject.libs.plugins.hilt.get().pluginId) {
+    //     hilt.enableAggregatingTask = true
```
sergei-lapin, line 98: it appears to be [enabled by default](https://github.com/google/dagger/blob/master/java/dagger/hilt/android/plugin/main/src/main/kotlin/dagger/hilt/android/plugin/HiltExtension.kt)

chrisbanes: Ah that must have changed 🤔

.gitignore
```diff
@@ -38,3 +38,7 @@ org.eclipse.buildship.core.prefs
 .classpath
 .project
 bin/
+
+# napt triggers
+# https://github.com/sergei-lapin/napt
+NaptTrigger.java
```
Goooler, line 44: We can put generated `NaptTrigger.java` files into proper packages, and [disabling NaptTrigger generation](https://github.com/sergei-lapin/napt#disabling-napttrigger-generation).

chrisbanes: We can't disable the triggers as we don't have any Java sources. The triggers ensure that napt runs on Kotlin-only projects.

Goooler: I meant we can move the triggers into java sourceSets and disable `generateNaptTrigger`. 

image
image

chrisbanes: Yeah we could, but I quite like that they're not polluting the 'actual' package we use. As they're generated code, it doesn't really matter imo. WDYT?

Goooler: No problem.

PR #1002, opened 2022-11-12
Enable gradle-home-cache-cleanup on gradle-build-action

Goooler:
https://github.com/gradle/gradle-build-action/releases/tag/v2.3.0
https://github.com/gradle/gradle-build-action/blob/main/README.md#removing-unused-files-from-gradle-user-home-before-saving-to-cache


Goooler:
What are the issues of 0296b68537b30caa4fc1be320c48c510df3a3ef7 due to the caches? Can we revert that commit?

PR #1012, opened 2022-11-27
Apply Ktlint trailing comma rules

Goooler:
~~Only apply `ij_kotlin_imports_layout` rule cause there are some issues with trailing comma rules, we'd better to migrate Spotless to Kotlinter.~~


chrisbanes (approved):

Goooler:
I've just applied trailing comma rules here.

PR #1016, opened 2022-11-28
Automate baseline profile generation

chrisbanes:
This PR automates baseline profile generation, by running our Macrobenchmark tests on an Android emulator, and then committing the updated baseline profile back to the repository.

If you'd like to know more about baseline profiles, I recommend looking through the [documentation](https://developer.android.com/topic/performance/baselineprofiles/overview).

#### Trigger
This workflow does not try to avoid commit cycles, whereby the workflow commits a new baseline profile, which then triggers the action to run again, which commits a new profile... repeat. Instead, it is designed to run on a schedule where it doesn't really matter if another load of workflows are triggered. Currently, it is set to run nightly.

#### Runner
It uses the more commonly used [reactivecircus/android-emulator-runner](https://github.com/ReactiveCircus/android-emulator-runner) to run the emulator. I would have liked to use [Gradle Managed Devices](https://developer.android.com/studio/test/gradle-managed-devices), but I couldn't get it to work at all on GitHub Actions.

However, using this runner seemed to break the automatic copy of the baseline profile from the emulator to the `build dir`. This is why there's a manual `adb pull` in the workflow.

### DIFF
--- a/.github/workflows/baseline-profile.yml
+++ b/.github/workflows/baseline-profile.yml
@@ -1,0 +1,85 @@
+name: Baseline profile generation
+
+on:
+  # every day at 00:43
+  schedule:
+    - cron: '43 0 * * *'
+  push: # remove this once verified that it runs!
+
+jobs:
+  baseline-profile:
+    runs-on: macos-latest
+    timeout-minutes: 45
+    env:
+      TERM: dumb
+      ORG_GRADLE_PROJECT_TIVI_TMDB_API_KEY: ${{ secrets.ORG_GRADLE_PROJECT_TIVI_TMDB_API_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/baseline-profile.yml
```diff
@@ -0,0 +1,85 @@
+name: Baseline profile generation
+
+on:
+  # every day at 00:43
+  schedule:
```
akshay253101, line 5: instead of schedule please set only run on push of release branch, and auto-push commit in same branch

chrisbanes: I don't have a release branch. If you have one, then yes that would work.

PR #1105, opened 2023-02-11
Migrate to kotlin-inject

### LINKED ISSUE #1096: Migrate to KMP compatible dependency injection
Probably to Koin, but I'm open-minded.

chrisbanes:
Using @evant's https://github.com/evant/kotlin-inject.
- [x] Add `SettingsActivity` component
- [x] Migrate all Compose screens
- [x] Fix WorkManager workers
- [x] Fix tests
- [x] Look at naming of `Module` classes.
- [x] Fix bad migration of `Provider`s (accidentally migrated them to `Lazy`).
- [x] Remove all traces of Hilt and JSR-330

Fixes #1096


api/tmdb/build.gradle.kts
```diff
@@ -17,7 +17,7 @@
 
 plugins {
     id("kotlin")
-    alias(libs.plugins.kotlin.kapt)
+    alias(libs.plugins.ksp)
```
dturner, line 20: Did you notice much build time improvement after making this change?

chrisbanes: Maybe? I don't have any numbers to back that up though. There's no more Java being compiled at all, which definitely helps. The goal was to support KMP though.

PR #1135, opened 2023-03-02
Migrate to moviebaseapp/tmdb-api

### LINKED ISSUE #1092: Migrate from tmdb-java

chrisbanes:
Fixes #1092


data/legacy/src/main/java/app/tivi/data/mappers/TmdbShowDetailToTiviShow.kt
```diff
@@ -16,15 +16,19 @@
...
-    override suspend fun map(from: BaseTvShow) = TiviShow(
+class TmdbShowDetailToTiviShow : Mapper<TmdbShowDetail, TiviShow> {
+    override suspend fun map(from: TmdbShowDetail) = TiviShow(
         tmdbId = from.id,
+        imdbId = from.externalIds?.imdbId,
         title = from.name,
         summary = from.overview,
+        homepage = from.homepage,
+        network = from.networks.firstOrNull()?.name,
+        networkLogoPath = null, // FIXME from.networks.firstOrNull()?.logo_path,
```
chrisbanes, line 32: Fixed upstream: https://github.com/MoviebaseApp/tmdb-api/pull/16

api/tmdb/src/main/java/app/tivi/tmdb/TmdbManager.kt
```diff
@@ -14,47 +14,46 @@
...
     suspend fun refreshConfiguration() {
-        try {
-            val response = withContext(dispatchers.io) {
-                tmdbClient.configurationService().configuration().awaitResponse()
-            }
-            onConfigurationLoaded(response.body()!!)
-        } catch (t: Throwable) {
-            // TODO
-        }
+//        try {
```
chrisbanes, line 38: Upstream PR: https://github.com/MoviebaseApp/tmdb-api/pull/18

data/relatedshows/src/main/java/app/tivi/data/relatedshows/TmdbRelatedShowsDataSourceImpl.kt
```diff
@@ -1,52 +0,0 @@
-/*
```
chrisbanes, line 1: Will un-remove this file once https://github.com/MoviebaseApp/tmdb-api/pull/17 is out

chrisnkrueger:
Two things to add in the TMDB lib:
- Retrying failed requests (https://ktor.io/docs/client-retry.html)
- Error handling for not found TMDB resources by a show ID
- 


For TMDB, there isn't a request limit anymore. So we don't need an extra retry mechanism. See https://developers.themoviedb.org/3/getting-started/request-rate-limiting

However, you may want to handle exceptions from the ktor request (ResponseException).

chrisbanes:
> For TMDB, there isn't a request limit anymore. So we don't need an extra retry mechanism. See https://developers.themoviedb.org/3/getting-started/request-rate-limiting

Nice. The retry handling was more for Trakt iirc.

Another thing I was thinking about was re-using my OkHttp client. Currently the creation of the Ktor client is private in Tmdb3, so I can't set the http engine. Maybe we could add a `(userConfig: HttpClientConfig) -> HttpClient` constructor param?

chrisbanes:
> For the entire OkHttp engine support, we need to open the library more. We can pass your own HTTP client instance and add the TMDB config on top.

Not necessarily. We can just rely on the implicit engine discovery (https://ktor.io/docs/http-client-engines.html#default), which is what I'm using right now. The important part is being able to configure the client underneath, through the `engine` [block](https://ktor.io/docs/http-client-engines.html#configure).

It _should_ work with https://github.com/MoviebaseApp/tmdb-api/pull/27 like so (although I haven't tested it):

```
val tmdb = Tmdb3(apiKey) {
    engine {
        if (this is OkHttpConfig) {
            //  okhttp stuff
        } else ...
    }
}
```

chrisnkrueger:
Yes, this should work. It feels a bit risky to rely on the selected engine from the current source set. Ktor looks for the first engine:
```
public actual fun HttpClient(
    block: HttpClientConfig<*>.() -> Unit
): HttpClient = engines.firstOrNull()?.let { HttpClient(it, block) } 
```

I added the following config MoviebaseApp/tmdb-api#28. I haven't tested the code, but it should work:
```
val tmdb = Tmdb3 {
            httpClient(OkHttp) {
               engine {
                    // for own client
               }
            }

            httpClient {
                // for custom config
            }
        }

```

PR #1190, opened 2023-04-01
Fix crash for failing TMDB configuration request

Tilizar:
Fix crash when TMDB configuration request fails somehow (timeout, http error, no internet, etc.)


chrisbanes (approved):
Thanks

PR #1218, opened 2023-04-27
Add ktlint to libs.versions.toml

mikescamell:
Much like the trick with the Compose Compiler we can do the same with ktlint to allow Renovate to update.


chrisbanes (approved):
Cheers Mike! Hope you're well

PR #1247, opened 2023-05-10
Add Gradle convention plugins

### LINKED ISSUE #1243: Use Gradle convention plugins to share build logic
Tivi Gradle modules share some common build logic. This is either done by duplicating it in every module (e.g. applying the cache fix plugin) or using the `allprojects` block.

Gradle's recommended way to do this is to use [convention plugins](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html). The idea is to have plugins like `app.tivi.android.library`, `app.tivi.android.application` (and apply these instead of the `com.android.library` and `com.android.application`). These plugins can then configure the common build logic. If needed, the modules can override these defaults (which is trickier to do with the `allprojects` block).

One downside with this approach is that you might forget to apply the convention plugins and use the standard Android plugins instead. If this is really an issue, the `allprojects` block can be used to check that every module applies one of the convention plugins.

Here are a few examples using it
https://github.com/android/nowinandroid/tree/main/build-logic
https://github.com/dropbox/focus/blob/main/settings.gradle#L2
https://github.com/jjohannes/gradle-project-setup-howto/blob/main/settings.gradle.kts#L2

erikghonyan:
Tivi Gradle modules share some common build logic. This is either done by duplicating it in every module (e.g. applying the cache fix plugin) or using the allprojects block.

Gradle's recommended way to do this is to use [convention plugins](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html). The idea is to have plugins like app.tivi.android.library, app.tivi.android.application (and apply these instead of the com.android.library and com.android.application). These plugins can then configure the common build logic. If needed, the modules can override these defaults (which is trickier to do with the allprojects block).

Here are a few examples using it
https://github.com/android/nowinandroid/tree/main/build-logic https://github.com/dropbox/focus/blob/main/settings.gradle#L2 https://github.com/jjohannes/gradle-project-setup-howto/blob/main/settings.gradle.kts#L2

Fixes #1243

### DIFF
--- a/android-app/app/build.gradle.kts
+++ b/android-app/app/build.gradle.kts
@@ -18,7 +18,7 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
 
 plugins {
-    alias(libs.plugins.android.application)
+    id("app.tivi.android.application")
     alias(libs.plugins.kotlin.android)
     alias(libs.plugins.cacheFixPlugin)
     alias(libs.plugins.ksp)
--- a/android-app/benchmark/build.gradle.kts
+++ b/android-app/benchmark/build.gradle.kts
@@ -18,7 +18,7 @@
 import com.android.build.api.dsl.ManagedVirtualDevice
 
 plugins {
-    alias(libs.plugins.android.test)
+    id("app.tivi.android.test")
     alias(libs.plugins.kotlin.android)
 }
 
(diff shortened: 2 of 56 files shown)


chrisbanes:
I'm going to go ahead and land this as a first step, as I have a use-case to share some more setup.

erikghonyan:
Sounds great! I was just about to push another commit to move the cache fix plugin, I'll open a new PR for it 🙂

PR #1262, opened 2023-05-16
Apply cache-fix plugin from convention plugins

erikghonyan:


chrisbanes (approved):
Ha, I just did the same thing locally. Thanks!

gradle/libs.versions.toml
```diff
@@ -28,7 +29,7 @@ android-application = { id = "com.android.application", version.ref = "agp" }
 android-library = { id = "com.android.library", version.ref = "agp" }
 android-lint = { id = "com.android.lint", version.ref = "agp" }
 android-test = { id = "com.android.test", version.ref = "agp" }
-cacheFixPlugin = { id = "org.gradle.android.cache-fix", version = "2.7.1" }
+cacheFixPlugin = { id = "org.gradle.android.cache-fix", version.ref = "cache-fix-plugin" }
```
erikghonyan, line 32: There's a bit of duplication here, the convention plugins use this same ID. I can also use the version catalog in the convention plugins to read its ID, but I think eventually we should be able to remove these plugins from here

erikghonyan:
Heh, sorry that I couldn't send this earlier 🙂

gradle/libs.versions.toml
erikghonyan: Hmm, or maybe I'm wrong. Let me clean this up a bit, I actually don't need that `compileOnly` dependency on this plugin

erikghonyan: Ok, I removed the dependency on the plugin. I still have the plugin id duplication, I think it's fine for now, we do it for the android application, library and multiplatform plugins as well. I want to play around with it a bit more to find a more elegant way, but otherwise I'll send a separate PR to use the version catalog. Will look something like this

```kotlin
        with(target) {
            val libs = extensions.getByType<VersionCatalogsExtension>().named("libs")

            with(pluginManager) {
                apply(libs.findPlugin("android-application").get().get().pluginId)
                apply(libs.findPlugin("cacheFixPlugin").get().get().pluginId)
            }
        }

```

erikghonyan: Ok, I think I found a way to avoid the plugin ID duplication 🙂 But we can go ahead and merge this PR, I'm waiting for some answers on that solution, I'll follow up with another PR

PR #1288, opened 2023-05-26
Migrate to Circuit

chrisbanes:


common/ui/circuit-overlay/src/main/kotlin/app/tivi/overlays/BottomSheetOverlay.kt
```diff
@@ -0,0 +1,95 @@
...
+import androidx.compose.ui.unit.Dp
+import com.slack.circuit.foundation.CircuitContent
+import com.slack.circuit.overlay.Overlay
+import com.slack.circuit.overlay.OverlayHost
+import com.slack.circuit.overlay.OverlayNavigator
+import com.slack.circuit.runtime.Screen
+import kotlinx.coroutines.launch
+
+@OptIn(ExperimentalMaterial3Api::class)
+class BottomSheetOverlay<Model : Any, Result : Any>(
```
ZacSweers, line 40: Do you think we should offer a batteries-included component like this? Definitely know there's like 4 different implementations of this same idea floating around different repos at the moment heh

common/ui/circuit-overlay/src/main/kotlin/app/tivi/overlays/BottomSheetOverlay.kt
```diff
@@ -0,0 +1,95 @@
...
+            scrimColor = if (scrimColor.isSpecified) scrimColor else BottomSheetDefaults.ScrimColor,
+            sheetState = sheetState,
+            onDismissRequest = { navigator.finish(onDismiss()) },
+        )
+
+        LaunchedEffect(Unit) { sheetState.show() }
+    }
+}
+
+suspend fun OverlayHost.showInBottomSheet(
```
ZacSweers, line 87: Nice

common/ui/circuit-overlay/src/main/kotlin/app/tivi/overlays/BottomSheetOverlay.kt
```diff
@@ -0,0 +1,95 @@
...
+    screen: Screen,
+): Unit = show(
+    BottomSheetOverlay(Unit, {}) { _, _ ->
+        // We want to use `onNavEvent` here to finish the overlay but we're blocked by
+        // https://github.com/slackhq/circuit/issues/653
```
ZacSweers, line 92: 👀

common/ui/screens/build.gradle.kts
ZacSweers, line 1: Defining all your screens separately's an interesting pattern. How do you like it vs co-locating them with the presenters/ui?

common/ui/screens/src/commonMain/kotlin/app/tivi/screens/Screens.kt
```diff
@@ -0,0 +1,76 @@
...
+    )
+}
+
+@CommonParcelize
+object TrendingShowsScreen : TiviScreen(name = "TrendingShows()")
+
+@CommonParcelize
+object UpNextScreen : TiviScreen(name = "UpNext()")
+
+abstract class TiviScreen(val name: String) : Screen {
```
ZacSweers, line 74: Out of curiosity, what's the benefit of defining a name string vs using the screen simple name? We do the latter in slack for analytics

common/ui/screens/src/commonMain/kotlin/app/tivi/screens/Screens.kt
```diff
@@ -0,0 +1,76 @@
...
+    open val arguments: Map<String, *>? = null
```
ZacSweers, line 75: I'd be curious to learn more about this!

ui/account/src/main/java/app/tivi/account/AccountComponent.kt
```diff
@@ -0,0 +1,35 @@
...
+interface AccountComponent {
+    @IntoSet
+    @Provides
+    @ApplicationScope
+    fun bindAccountPresenterFactory(factory: AccountUiPresenterFactory): Presenter.Factory = factory
+
+    @IntoSet
+    @Provides
+    @ApplicationScope
+    fun bindAccountUiFactory(factory: AccountUiFactory): Ui.Factory = factory
```
ZacSweers, line 34: Do you think there's anything we could do to make these kind of stuff easier to use in other DI systems like kotlin-inject? We built anvil support mostly because that's just what we use, but open to supporting other stuff if there's stuff you think is reusable

ui/account/src/main/java/app/tivi/account/AccountPresenter.kt
```diff
@@ -21,28 +21,47 @@ import androidx.compose.runtime.LaunchedEffect
...
+class AccountUiPresenterFactory(
+    private val presenterFactory: (Navigator) -> AccountPresenter,
+) : Presenter.Factory {
+    override fun create(screen: Screen, navigator: Navigator, context: CircuitContext): Presenter<*>? {
+        return when (screen) {
+            is AccountScreen -> presenterFactory(navigator)
+            else -> null
+        }
+    }
+}
```
ZacSweers, line 51: For cases not using a code gen tool, anything we could do to make these simpler? Maybe a `ScreenPresenterFactory(Screen,Function)` API that dries some of this up?

ui/discover/src/main/java/app/tivi/home/discover/Discover.kt
```diff
@@ -76,82 +75,80 @@ import app.tivi.common.compose.bodyWidth
...
-    openTrendingShows: () -> Unit,
-    openPopularShows: () -> Unit,
-    openRecommendedShows: () -> Unit,
-    openShowDetails: (showId: Long, seasonId: Long?, episodeId: Long?) -> Unit,
-    openUser: () -> Unit,
+    state: DiscoverUiState,
+    modifier: Modifier = Modifier,
 ) {
-    val viewState = viewModel.presenter()
+    val scope = rememberCoroutineScope()
```
ZacSweers, line 114: You may wanna check out StableCoroutineScope: https://github.com/slackhq/circuit/blob/main/samples/star/src/main/kotlin/com/slack/circuit/star/ui/StableCoroutineScope.kt

ui/account/src/main/java/app/tivi/account/AccountComponent.kt
ZacSweers: Another option is you could group these together into one composite factory, and possibly even cover multiple screens within it

ui/episode/track/src/main/java/app/tivi/episode/track/EpisodeTrackPresenter.kt
```diff
@@ -48,17 +51,30 @@ import me.tatarka.inject.annotations.Assisted
...
-class EpisodeTrackViewModel(
-    @Assisted savedStateHandle: SavedStateHandle,
+class EpisodeTrackUiPresenterFactory(
+    private val presenterFactory: (EpisodeTrackScreen, Navigator) -> EpisodeTrackPresenter,
+) : Presenter.Factory {
+    override fun create(
+        screen: Screen,
+        navigator: Navigator,
+        context: CircuitContext,
+    ): Presenter<*>? = when (screen) {
```
ZacSweers, line 61: Side not - these are `fun interface` so could it be simpler to use the function syntax?

ZacSweers (commented):
This is really cool! Bunch of questions for ya about things that we could maybe do better on the Circuit side

ui/episode/track/src/main/java/app/tivi/episode/track/EpisodeTrackPresenter.kt
ZacSweers: ehh nevermind I guess it's not possible to use class delegation with a function input. Tried it locally without success

common/ui/circuit-overlay/src/main/kotlin/app/tivi/overlays/BottomSheetOverlay.kt
chrisbanes: Yeah I was thinking the same thing. Probably worth keeping as a separate artifact. It also gets tricky with Jetpack Compose vs Compose Multiplatform.

common/ui/screens/build.gradle.kts
chrisbanes: If you co-locate them with the presenters, how do you navigate to another screen?

I didn't want screen modules depending on other screen modules, and basically I see these classes as Intents.

Having them globally available allows the only coupling to be these objects.

ZacSweers: fair point!

common/ui/screens/src/commonMain/kotlin/app/tivi/screens/Screens.kt
chrisbanes: I wanted to avoid weirdness with obfuscation. I also have existing labels and wanted these to match for consistency.

common/ui/screens/src/commonMain/kotlin/app/tivi/screens/Screens.kt
chrisbanes: These are solely there for analytics. These gets pushed up to Crashlytics along with screen views.

ui/account/src/main/java/app/tivi/account/AccountComponent.kt
chrisbanes: Personally I didn't have a problem with writing these. 

kotlin-inject makes creating the assisted injection factory super simple, which helped a lot.

I'm not sure how much of this stuff could be transferable across DI systems tbh.

ui/account/src/main/java/app/tivi/account/AccountPresenter.kt
chrisbanes: I thought about writing exactly that, but the only thing you're really saving is the `when`.

You still need a class somewhere to inject the assisted factory. I guess you could move that to the component/module instead, and just call a `ScreenPresenterFactory()`.

PR #1301, opened 2023-06-04
Move Spotless configuration to convention plugins

erikghonyan:


gradle/build-logic/convention/src/main/java/app/tivi/gradle/Spotless.kt
```diff
@@ -0,0 +1,54 @@
...
+
+import com.diffplug.gradle.spotless.SpotlessExtension
+import org.gradle.api.Project
+import org.gradle.api.artifacts.VersionCatalogsExtension
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.getByType
+
+fun Project.configureSpotless() {
+    val libs = extensions.getByType<VersionCatalogsExtension>().named("libs")
+    val ktlintVersion = libs.findVersion("ktlint").get().requiredVersion
```
erikghonyan, line 14: To avoid this we could also remove our version override and use what comes as default with Spotless

chrisbanes (approved):
Looks good, thanks!

PR #1688, opened 2023-12-28
Migrate to enableEdgeToEdge() function

Omico:


chrisbanes:
The `enableEdgeToEdge()` function is a convenience function, which ends up needing more code (#1689) to achieve the same thing.

I'm not going to merge this or #1689 as there is no real benefit, only more code.

Thanks for spending the time though!

Omico:
> I'm not going to merge this or https://github.com/chrisbanes/tivi/pull/1689 as there is no real benefit, only more code.

For #1689, The status bar won't change to the light style after manually changing to the dark theme. So we have to add these codes.

chrisbanes (approved):

itsandreramon:
What's the reason you changed your mind?

chrisbanes:
I want to refactor the code, but this does fix an issue with button navigation bars. I personally don't use it, but that doesn't mean it shouldn't be fixed.

chrisbanes:
Thanks for the contribution @Omico!

PR #1689, opened 2023-12-28
Handle system bar style when the theme changed

Omico:


chrisbanes:
See https://github.com/chrisbanes/tivi/pull/1688#issuecomment-1871193665

PR #1691, opened 2024-01-01
Migrate to Coil 3

chrisbanes:
Fully working, just want to test it more before merging.


chrisbanes:
FYI @colinrtwhite in case you're interested

common/imageloading/build.gradle.kts
```diff
@@ -24,7 +23,8 @@ kotlin {
 
         implementation(libs.kotlininject.runtime)
 
-        api(libs.imageloader)
+        api(libs.coil.core)
+        api(libs.coil.network)
```
colinrtwhite, line 27: You'll need to import a Ktor engine for each platform as well. Here's what I use: https://github.com/coil-kt/coil/blob/main/samples/shared/build.gradle.kts#L44

common/imageloading/src/commonMain/kotlin/app/tivi/common/imageloading/ImageLoader.kt
```diff
@@ -0,0 +1,67 @@
...
+      add(NetworkFetcher.Factory())
+      interceptors.forEach(::add)
+    }
+    .memoryCache {
+      MemoryCache.Builder()
+        .maxSizePercent(context, percent = 0.25)
+        .build()
+    }
+    .diskCache {
+      DiskCache.Builder()
```
colinrtwhite, line 37: If you plan to run this on the web you should have this return a null `DiskCache` on JS like [here](https://github.com/coil-kt/coil/blob/main/samples/shared/src/commonMain/kotlin/sample/common/imageLoader.common.kt#L41).

It's not possible to write to the file system on the web so trying to write to it will (currently) throw.

common/imageloading/build.gradle.kts
chrisbanes: There's no huge need here, as the Ktor engines are already on the app classpath.

common/imageloading/src/commonMain/kotlin/app/tivi/common/imageloading/ImageLoader.kt
chrisbanes: No JS target as yet (and I doubt I'll ever bother).

PR #1918, opened 2024-06-27
Move Circuit events to be ran in processor

chrisbanes:


LloydBlv:
I wonder what's the logic behind this change @chrisbanes ?

chrisbanes:
Just something I'm experimenting with. It gives more control over running Coroutines from user events.

The downside is that long running tasks could block events which require immediate processing (go to next screen, etc)

LloydBlv:
> Just something I'm experimenting with. It gives more control over running Coroutines from user events.
> 
> The downside is that long running tasks could block events which require immediate processing (go to next screen, etc)

Thanks @chrisbanes , I had an issue using Circuit sometime ago, the values captured inside eventSink looks like are not up to date, do you think this will cover those cases?

chrisbanes:
> The values captured inside eventSink looks like are not up to date, do you think this will cover those cases?

Captured values will be an issue either way. This would just defer the work to be queued a bit later.

LloydBlv:
> > The values captured inside eventSink looks like are not up to date, do you think this will cover those cases?
> 
> Captured values will be an issue either way. This would just defer the work to be queued a bit later.

I see, Do you have any solution on top of your mind for captured values? Does having remember{} fields help?

chrisbanes:
> I see, Do you have any solution on top of your mind for captured values? Does having remember{} fields help?

You want to reference state holders, rather than values. Have a look at using [rememberUpdatedState](https://developer.android.com/develop/ui/compose/side-effects#rememberupdatedstate).

LloydBlv:
> > I see, Do you have any solution on top of your mind for captured values? Does having remember{} fields help?
> 
> You want to reference state holders, rather than values. Have a look at using [rememberUpdatedState](https://developer.android.com/develop/ui/compose/side-effects#rememberupdatedstate).

Thanks, Do you think it's worth being mentioned in the Circuit docs? Or maybe this should come as Compose domain knowledge already :)

chrisbanes:
Maybe? Worth raising an issue to discuss anyway.

mslalith:
The draft looks cool! I was also looking into ways to not create CoroutineScope ourselves. So now instead of creating CoroutineScope ourselves, now the scope is taken from LaunchedEffect.

I initiated a discussion for this at https://github.com/slackhq/circuit/discussions/1634

PR #1979, opened 2024-08-05
Fix edge-to-edge on three-button nav bar

Goooler:
| Before | After |
|--------|--------|
| before | after |


chrisbanes (commented):
I'm not sure on this. 

Tivi does draw provide any content protection for the navigation bar, as the bottom navigation is floating above. The comments mentioned in the linked issue are correct about the buttons needing to be accessible, therefore the current behavior is likely the correct one.

Goooler:
The background color should be drawn on navigation bar, it works for gesture navigation but 3-button.


https://github.com/user-attachments/assets/5d029477-9b8f-4f7b-96f8-0a882d91614d

chrisbanes:
Thanks! I'll try it now and see what I think

